PR #129, opened 2015-01-30
Ability to download XML and TSV report.

vivekiitkgp:


vivekiitkgp:
Will need to rework the test suite a bit.

yannickwurm:
Thanks - Heads up that Travis build is failing. 

Can this be finalised so it can be deployed to `--pre` gem later today so we can announce the whole prerelease to mailing list before the weekend?

vivekiitkgp:
@yannickwurm Do you like this?

yeban:
@vivekiitkgp I think we should provide two links in the sidebar:

    TSV Report: [standard] [full]

And detail the columns [standard] will contain in a BS tooltip.

If you want to provide TSV customisation, you could add a 3rd link (if space allows) - [custom] which opens the modal in its current form sans the checkboxes. I'm up for custom TSV, but I would like to click on the name of columns rather than type them out in an input box.

vivekiitkgp:
@yannickwurm I would also like to have few words from you before we modify the current implementation. An early feedback will reduce total invested time.

yannickwurm:
Hey @vivekiitkgp - give me a screenshot please. 
 1. Naming:  ~~TSV Report:~~ -> "Tab-delimited table:" (most humans haven't heard of TSV). 
 2. We do not want/need customisation - noone will ever use it; and it adds implementation/maintenance overhead (options can differ btw blast versions...)

vivekiitkgp:
@yannickwurm Here are possible use case scenarios and relevant screenshots.

screenshot-localhost 4567 2015-02-01 08-05-36
screenshot-localhost 4567 2015-02-01 08-05-05
screenshot-localhost 4567 2015-02-01 08-04-29

yannickwurm:
Whoa - this is waaaaay overkill.=20

No modals. No extra unnecessary clicking.



Just do the simplest thing possible:=20
Just replace "TSV report (coming soon)"  with "Tab delimited report =
[Standard] [Full]"





> On 1 Feb 2015, at 02:36, Vivek Rai <notifications@github.com> wrote:
>=20
> @yannickwurm Here are possible use case scenarios and relevant =
screenshots.

yeban:
Please rebase on top of latest master. A lot has changed in the last 3 commits and there will be merge conflicts.

spec/ss_sample_blast_no_hits.asn
```diff
@@ -0,0 +1,130 @@
+Blast4-archive ::= {
```
yeban: Archive format (11) is binary, I believe. How come using text file for test suite?

vivekiitkgp: It (outfmt 11) is a readable text file.

lib/sequenceserver.rb
```diff
@@ -16,6 +16,21 @@ module SequenceServer
   # Use a fixed minimum version of BLAST+
   MINIMUM_BLAST_VERSION           = '2.2.30+'
 
+  OUTFMT = {
```
yeban: This should be in BLAST.

yeban: Together with the reference link [1] below.

lib/sequenceserver.rb
```diff
@@ -373,6 +388,26 @@ def tuple?(data)
...
+      rfile  = File.join(Dir.tmpdir, params[:report])
+
+      file_name = "sequenceserver_report_#{Time.now.strftime('%H%M')}.#{format}"
+      file = Tempfile.new file_name
+
+      case format
+      when 'xml'
+        file.write(BLAST.blast_formatter(rfile, OUTFMT['xml']))
+      when 'tsv'
+        file.write(BLAST.blast_formatter(rfile, OUTFMT['tsv'], specifiers))
```
yeban: Commented TSV could be more helpful.

lib/sequenceserver/blast.rb
```diff
@@ -104,6 +103,10 @@ def pre_process(params)
         params[:sequence].strip! unless params[:sequence].nil?
       end
 
+      def blast_formatter(rfile_path, ofmt, specifiers = nil)
+        `blast_formatter -archive '#{rfile_path}' -outfmt '#{ofmt} #{specifiers}'`
```
yeban: `specifiers` is technically a user input. Thus should be verified, and used only if applicable. That is, only for TSV, and only after checking its validity.

lib/sequenceserver.rb
```diff
@@ -373,6 +388,26 @@ def tuple?(data)
       end
     end
 
+    get '/get_report/' do
+      format = params[:format]
+      specifiers = params[:specifiers]
+
+      rfile  = File.join(Dir.tmpdir, params[:report])
```
yeban: Verify it's existence before using it.

spec/ss_sample_blast_no_hits.asn
yeban: How?

lib/sequenceserver/blast.rb
vivekiitkgp: A good thing about specifiers here is that wrong values are happily ignored. So if I pass `specifiers  = "qframe random none"`, then output will contain only of `qframe`.

yeban:
Regarding failing test suite, do note that rubocop is a part of the test suite now.

spec/ss_sample_blast_no_hits.asn
vivekiitkgp: Binary ASN.1 is `-outfmt 9`

lib/sequenceserver/blast.rb
yeban: Fair enough. Add a note in the source code though.

What's the behaviour with XML?

vivekiitkgp: Doesn't affect it.

yeban: Fine then. Let BLAST deal with incorrect values (if at all). Do make a note of the behaviour.

yeban:
Keep the controllers as thin as possible. Like, `post '/'`. I think all the logic of `get_report/` can be moved to `BLAST.blast_formatter`.

yeban:
And should call the method `format` instead of `blast_formatter`.

lib/sequenceserver.rb
```diff
@@ -373,6 +388,26 @@ def tuple?(data)
...
+
+      file_name = "sequenceserver_report_#{Time.now.strftime('%H%M')}.#{format}"
+      file = Tempfile.new file_name
+
+      case format
+      when 'xml'
+        file.write(BLAST.blast_formatter(rfile, OUTFMT['xml']))
```
yeban: Use `Hash#fetch` instead of `[]`. Rescue `KeyError`, and re-raise it as 400 Bad Request.

yeban:
Need to add logging for debug mode.

yeban:
Safari add `.txt` extension to XML download. Will need to use appropriate mime type.

yeban:
Need to remove `[]` from around the links. Should use `|` instead of `/` as the separator like everywhere else in the HTML report.

yeban:
Should inform users what columns standard TSV will contain. Via BS tooltip, perhaps.

PR #131, opened 2015-02-01
[RFC] Kablammo

vivekiitkgp:
@yannickwurm Please review.

Currently, the behavior is as follows:
* Users open the report viewer and is taken to the intro page.
* They select the load file option and select the generated xml file from option.
* Voila!

This commit is based on the changes pending in other PR (which will be finalized soon)


public/ext/kablammo/data/sequenceserver_blast_result20150201-27830-1cjljf8.xml
```diff
@@ -0,0 +1,2558 @@
+<?xml version="1.0"?>
```
yeban: Is this a sample result or committed by mistake? This needs to be removed. It's confusing.

vivekiitkgp: By mistake.

yannickwurm:
Re the Kablammo report - the user should not have to load/download XML file. A single click to open the report in new window should be plenty. =

yeban:
Also, see if you can't include Kablammo as a git submodule.

yannickwurm:
@yeban  please push to pre make announcement on mailing list without this.

yeban:
@yannickwurm I'm waiting for #129.

vivekiitkgp:
@yannickwurm Kablammo doesn't provide any API as such. For that, I think we need to add custom Javascript to its main page. This will probably prevent us from directly using it as a submodule.

Not sure about it.

yeban:
Worth asking how to go about that on their issue tracker?

Issue #146, opened 2015-02-27
Feature Request: Automatically open the Localhost link in the default web browser 


IsmailM:
Instead of writing go to http://localhost:4567, automatically open the link in the default browser. This can be quite easily done as follows:

Mac: `open http://localhost:4567`

Linux: `xdg-open http://localhost:4567`

yeban:
Send a PR?

Will have to take into account the fact xdg-open is not available by
default on all Linux distros. And I won't encourage SequenceServer trying
to install it.

-- Priyam

IsmailM:
Cool I'll try to put something together this weekend...

yannickwurm:
I suggest doing this *only* on mac.

And *only* if a user is logged in on a normal GUI. 

(ie not if only connections are via ssh/telnet)

IsmailM:
Any Ideas on how to check:

* if the user's machine has a GUI
* if the user is using SSH/telnet

I have created a stack overflow post on this [here](http://stackoverflow.com/questions/28797141/check-whether-host-supports-gui-and-check-if-user-is-sshing-telnet)

yeban:
RUBY_PLATFORM

-- Priyam

yeban:
Hey @IsmailM. Please could you suppress this behaviour in development mode? It's getting a bit annoying for me as I have to sometimes restart the server to load the changes but need not open a new window and BLAST each time.

PR #153, opened 2015-03-06
open the local server in the default browser

IsmailM:
PR for #146.

Opens up the local server in the default browser...

Doesn't open the default browser if the user is SSHing or if the machine does not have a GUI...

### DIFF
--- a/lib/sequenceserver.rb
+++ b/lib/sequenceserver.rb
@@ -63,6 +63,7 @@ module SequenceServer
         puts '** SequenceServer is ready.'
         puts "   Go to #{url} in your browser and start BLASTing!"
         puts '   Press CTRL+C to quit.'
+        open_default_browser(url)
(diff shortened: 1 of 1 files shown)


IsmailM:
Hey Priyam,

`RUBY_PLATFORM` seems to be slightly unreliable method of obtaining the OS (e.g. `RUBY_PLATFORM` return `java` when using JRuby, regardless of the OS)...

Do you think this is something we should address....

Alternatives include the `os` gem or [this method](http://stackoverflow.com/a/13586108)...

NB. There is also some code in the `bin/sequenceserver` that uses RUBY_PLATFORM...

yeban:
Works great on Mac. Thanks!

No need to worry about OS detection approach as of now. If someone is using JRuby, they will know to open the link in a browser as well.

PR #154, opened 2015-03-06
Show the GET/POST requests made by the server only in the development environment

IsmailM:
The Rack::CommonLogger shows all the GET/ POST etc. requests made by the server. Most normal people do not care about this information...

Thus, I think it would be best to only show this information when in the development environment and not in the production environment (i.e. only when using the `-D` command line argument)

Any thoughts?

It would also make it easier for ('new to bash') users to see how to close the local server...

Signed-off-by: IsmailM <ismail.moghul@gmail.com>

### DIFF
--- a/lib/sequenceserver.rb
+++ b/lib/sequenceserver.rb
@@ -55,7 +55,7 @@ module SequenceServer
       server = Thin::Server.new(config[:host],
                                 config[:port],
                                 :signals => false) do
-        use Rack::CommonLogger
+        use Rack::CommonLogger if SequenceServer.environment == 'development'
         run SequenceServer
       end
       server.silent = true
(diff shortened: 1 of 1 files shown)


lib/sequenceserver.rb
```diff
@@ -55,7 +55,7 @@ def run
       server = Thin::Server.new(config[:host],
                                 config[:port],
                                 :signals => false) do
-        use Rack::CommonLogger
+        use Rack::CommonLogger if SequenceServer.environment == 'development'
```
yeban: 1. Should make use of pre-defined convenience methods. See `SequenceServer.verbose?`.
2. Do not specify receiver unless required.

yeban:
I agree. I will merge as soon as you have made above changes.

IsmailM:
Hey Priyam,

I have made the requested change…

It seems that the receiver is required here (it fails otherwise, possible because it is in the Thin.new loop)...

PR #155, opened 2015-03-06
add a physical close button ('x') to the sequence-viewer modal

IsmailM:
This is not a necessary feature, but I think that it's a good idea to have a physical button that dismisses the modal for optimal user experience 

Users can still click outside the modal or press <kbd>esc</kbd> to dismiss the modal, but I don't think this would be glaringly obvious to everyone...

Thoughts...

Signed-off-by: IsmailM ismail.moghul@gmail.com

### DIFF
--- a/views/search.erb
+++ b/views/search.erb
@@ -406,6 +406,7 @@
         class="modal-content">
         <div
           class="modal-header">
+          <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
           <h3>View and download sequence</h3>
         </div>
 


yeban:
It's 1000 times easier to click anywhere outside the modal to dismiss it rather than clicking on a close button. I feel in the presence of a close button users will never discover the more optimal way. But I agree it can be a bit confusing the first time. If there has to be a close button, I would rather the text be 'Done' / 'Dismiss'.

Have you reviewed other modals? error modals and one that lists advanced params have only one button that dismisses the modal, no close button required. Then there's 'blast-ing' modal with a spinner that doesn't require any buttons. Any other modal that we need to consider?

And I would need opinion of others before deciding on this PR. /cc @yannickwurm @wolfgangrumpf.

wolfgangrumpf:
Having no button at all can also be frustrating, as you pointed out.  If it’s that much of a concern, add the “Done” button but also put in fine print someone at the bottom of the dialog “click anywhere or press ‘Done” when finished"




Cheers,


Dr. Wolfgang Rumpf
Research Scientist, Mathematical Medicine, The Research Institute at Nationwide Children's Hospital
Associate Professor, Dept. of Biotechnology, UMUC
wolfgang.rumpf@nationwidechildrens.org . wolfgang.rumpf@gmail.com
Mobile - (614) 638-6797 . Skype - wolfgang.rumpf
---------------------------------------------------------------------------------------------

IsmailM:
The fine print idea could work quite well...

Looking at other modals in SS; Most of the other modals have physical dismissal button in the modal footer (e.g. 'Got it!' ). 

Thus, this sequence viewer  and the BLAST modal are the only two modals with no physical dismissal button. (I also think the general error modal does not have a physical dismissal button)


On a side note, with the BLAST modal, there is no way of cancelling once you click on the BLAST button. Perhaps it would be an idea to have a cancel button (which sends a kill signal to BLAST process)...

Currently, I find that if I accidentally input the wrong parameters and want to restart BLAST, I have to refresh the page; and then manually restart the server to kill the BLAST process (which would be taking up most of my CPU/RAM)...

yeban:
I am happy to go with a 'Done' / 'Dismiss' button at bottom. No fine-print.

@IsmailM Would it be possible for you to update the PR against master branch?

yeban:
This was a good idea. I merged the PR into 1.0.x branch.

PR #156, opened 2015-03-06
make rows clickable on the HSPs overview table

IsmailM:
In my opinion, this could be a useful (User Experience) feature - It's easier to click on the row than move the cursor to the sequence ID and then click on it. 

Any thoughts...

Signed-off-by: IsmailM <ismail.moghul@gmail.com>


yeban:
Can still copy text from the row?

IsmailM:
Yep, it's very difficult to copy text from the row with this...

As such, I'm closing this PR....

:)

Issue #158, opened 2015-03-08
Cannot open http://0.0.0.0:4567 in Chrome


IsmailM:
I cannot open `http://0.0.0.0:4567` in the latest version of chrome (on yosemite); rather it opens up the Google search results of the query http://0.0.0.0:4567...

`http://localhost:4567` works fine...

Perhaps we should stick with `localhost:4567` if it is more reliable...

IsmailM:
There seems to be a quite a few Issues opened on Chromium with regards to this. I think they are planning on a fix... 

https://code.google.com/p/chromium/issues/detail?id=428046

Nevertheless, despite a fix was promised over a month ago; the problem is still here in Chrome Canary (which is two versions newer than the latest chrome)...

So, it may be some time before a fix actually comes into place...

yeban:
Agreed. Can you take care of this?

PR #161, opened 2015-03-09
Change the host default to localhost

IsmailM:
As requested in #158...

@yeban:
* Do you have a preference between `http://localhost:4567` & `http://127.0.0.1:4556`
* Secondly, should I remove the `check_host` [method](https://github.com/yannickwurm/sequenceserver/blob/master/lib/sequenceserver.rb#L153) completely or is still worthwhile checking if the host is `http://0.0.0.0`?

NB: you may need to clear the `host` value in your config file in order to use the new default.

### DIFF
--- a/lib/sequenceserver.rb
+++ b/lib/sequenceserver.rb
@@ -150,7 +150,7 @@ module SequenceServer
       raise NUM_THREADS_INCORRECT
     end
 
-    # Check and warn user if host is 0.0.0.0 (default).
+    # Check and warn user if host is 0.0.0.0.
     def check_host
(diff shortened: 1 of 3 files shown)


yeban:
Sorry, I should have been clear. We can't change the default host to localhost. Binding to localhost means you can connect to the server only from local computer. Binding to 0.0.0.0 means the server will accept incoming connections over the network as well.

I was thinking, should display localhost:4567 in case the host is one of localhost, 127.0.0.1, or 0.0.0.0. Show the value of host from config otherwise.

IsmailM:
Cool, I'll open a new PR in a bit...

PR #165, opened 2015-03-10
open localhost if host is '127.0.0.1' or '0.0.0.0'

IsmailM:
In response to #158...

SS should now show `localhost` in case the host is one of the follow: `localhost`, `127.0.0.1`, or `0.0.0.0`. 

Nevertheless, the Thin Server will still be opened at the configured host.

### DIFF
--- a/lib/sequenceserver.rb
+++ b/lib/sequenceserver.rb
@@ -185,11 +185,9 @@ module SequenceServer
     end
 
     def server_url
-      if config[:host] == '127.0.0.1' || config[:host] == '0.0.0.0'
-        "http://localhost:#{config[:port]}"
-      else
-        "http://#{config[:host]}:#{config[:port]}"
-      end
+      host = config[:host]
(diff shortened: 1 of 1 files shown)


lib/sequenceserver.rb
```diff
@@ -185,6 +184,14 @@ def assert_blast_installed_and_compatible
       fail BLAST_NOT_COMPATIBLE, version unless version >= MINIMUM_BLAST_VERSION
     end
 
+    def server_url
```
yeban: I would have written it like this:

```
host = config[:host]
host = 'localhost' if host == '127.0.0.1' || host == '0.0.0.0'
"http://#{host}:#{config[:port]}
```

IsmailM:
Thanks, that reads a lot better

Secondly, shall I change the variable in the `open_default_browser` method from `url` to `server_url` in order to make the whole file easier to read...

yeban:
Makes sense.

I would also prefer the name 'open_in_browser' for 'open_default_browser'.

PR #171, opened 2015-03-10
Responsive advanced param and BLAST button input group.

yeban:
As per @yannickwurm, @vivekiitkgp, @IsmailM 's feedback.

### DIFF
--- a/public/css/custom.css
+++ b/public/css/custom.css
@@ -276,6 +276,15 @@ a.disabled:hover {
 label[for="advanced"] {
   font-family: "Roboto";
   font-size: 16px;
+  display: table-cell;
+  width: 1%;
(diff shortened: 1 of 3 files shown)


vivekiitkgp:
Thanks for the changes @yeban. There's a small thing that would make this perfect. When the screen is resized, the tooltip on the BLAST button, being left aligned, is cut off from the window.

yeban:
Right. I'm guessing SequenceServer will be used mostly on small (sm), medium (md) and large (lg) devices. Not extra small (xs). It works well for those. I believe we should be fully responsive, but feel that can wait a bit. For instance, notifications need refining for xs as well (sm too, I think). I think I will open a new issue to track those.

PR #177, opened 2015-03-16
Make the SVGs responsive

### LINKED ISSUE #173: screen resize issue
There's a small transition period with snafu (e.g. at ~1 seconds and ~11 seconds)

https://www.youtube.com/watch?v=EJxfIclhxVA

IsmailM:
This makes the SVGs responsive (and as such it fixes #173), by redrawing the SVG on a browser resize...
<br>

NOTE: The first commit is an alternative solution, which calculated an aspect ratio to resize the SVG... However, I found that the text within the SVG was also resized. As a result, the SVG axis font would sometimes be too small (when decreasing browser width) or too big (when increasing browser width), which made the graphical overview chart look really weird... Thus I came up with this alternative solution...

What do you think?


yeban:
I can took a look into this Wednesday or later. Perhaps @vivekiitkgp will be able to advise in the mean time. He wrote the graphical overview.

public/js/jquery.t.js
```diff
@@ -16,6 +16,23 @@
         });
     };
 
+    var debounce = function(fn, timeout) {
```
vivekiitkgp: This should be `function (fn, timeout)`

public/js/jquery.t.js
```diff
@@ -16,6 +16,23 @@
...
+      var timeoutID = -1;
+      return function() {
```
vivekiitkgp: Same here.

public/js/jquery.t.js
```diff
@@ -16,6 +16,23 @@
...
+        if (timeoutID > -1) {
+          window.clearTimeout(timeoutID);
+        }
+        timeoutID = window.setTimeout(fn, timeout);
+      }
+    };
+   
+    var setupResponsiveness = function ($queryDiv, $graphDiv, index, howMany, opts)  {
+        var debounced_draw = debounce(function() {
```
vivekiitkgp: Same here.

public/js/jquery.t.js
```diff
@@ -348,6 +365,8 @@
             // Ensure clicking on 'rect' takes user to the relevant hit on all
             // browsers.
             setupClick($graphDiv);
+            // Redraw the graph on a browser resize...
+            setupResponsiveness($queryDiv, $graphDiv, index, howMany, opts, resize);
```
vivekiitkgp: `setupResponsiveness` doesn't take the `resize` parameter.

public/js/jquery.t.js
```diff
@@ -16,6 +16,23 @@
...
     };
 
+    var debounce = function(fn, timeout) {
+      var timeoutID = -1;
+      return function() {
+        if (timeoutID > -1) {
+          window.clearTimeout(timeoutID);
+        }
+        timeoutID = window.setTimeout(fn, timeout);
+      }
```
vivekiitkgp: Semicolon

public/js/jquery.t.js
```diff
@@ -340,7 +357,7 @@
             }
             // Bind listener events once all the graphical elements have
             // been drawn for first time.
-            if (index === 0) {
+            if (index === 0 && resize !== 1 ) {
```
vivekiitkgp: Can you document the `resize` parameter (as in, why it is important)? May be in `setupResponsiveness` or here itself.

vivekiitkgp:
@IsmailM I have added few comments. Once they are addressed, it is good to go.

public/js/jquery.t.js
```diff
@@ -16,6 +16,23 @@
...
+    var debounce = function(fn, timeout) {
+      var timeoutID = -1;
+      return function() {
+        if (timeoutID > -1) {
+          window.clearTimeout(timeoutID);
+        }
+        timeoutID = window.setTimeout(fn, timeout);
+      }
+    };
+
```
vivekiitkgp: Trailing spaces here.

IsmailM:
@vivekiitkgp 
Thanks for having a look... I've tidied it up...

vivekiitkgp:
@IsmailM Probably squash the last commit into the previous one? I feel that having a separate commit to fix minor things would add to the log noise, given that you can integrate those changes in the main commits.

It's not an issue per se, just a personal thought ;) Rest is cool.

PR #182, opened 2015-03-18
make the SVGs responsive

IsmailM:
Makes SVGs responsive

@vivekiitkgp


public/js/jquery.t.js
```diff
@@ -16,6 +16,24 @@
         });
     };
 
+    var setupResponsiveness = function ($queryDiv, $graphDiv, opts)  {
+        var debounce = function (fn, timeout) {
```
yeban: Doesn't http://underscorejs.org/#debounce serve your purpose? We have underscore, iirc.

public/js/jquery.t.js
```diff
@@ -348,6 +366,8 @@
             // Ensure clicking on 'rect' takes user to the relevant hit on all
             // browsers.
             setupClick($graphDiv);
+            // Redraw the graph on a browser resize...
+            setupResponsiveness($queryDiv, $graphDiv, opts);
```
yeban: Are you sure this doesn't attach the same handler several times to `window.resize`?

@vivekiitkgp Can you verify the case for `setupClick` as well?

IsmailM:
Have now updated the PR and have addressed all above points...

public/js/jquery.t.js
```diff
@@ -16,6 +16,17 @@
         });
     };
 
+    var setupResponsiveness = function ($queryDiv, $graphDiv, index, opts)  {
+        var debounced_draw = _.debounce(function () {
+            var shownHits = $queryDiv.find('.ghit > g').length;
+            $.graphIt($queryDiv, $graphDiv, shownHits, 0, opts);
```
yeban: I think it should be `index` instead of `0`. Otherwise I see a lot of errors raised by d3 on resize in the console if 'View more' button is present in the graph. I think s/0/index intuitively makes more sense too.

To reproduce - blastn first 50 lines of a scaffold to that genome so you get a lot of hits.

public/js/jquery.t.js
```diff
@@ -16,6 +16,17 @@
...
+        }, 125);
+        // Bind resize event only when the SVG is created for the first time.
+        if (index === 0) {
```
yeban: I wouldn't have raised this if I hadn't found the above bug:

I see `if (index === 0)` check at L356 as well (when your patch is applied). Is it not enough to call `setupResponsiveness` in the same conditional?

yeban:
Thanks for this @IsmailM. I like how simple the solution is. +1 for debounce and +1 for taking shownHits into account. I was wondering if resize should cause redraw only if browser width change. But is that info readily available? What do you think?

(@vivekiitkgp: can't the `graphit` API be improved there so that how many hits are shown is handled internally?)

vivekiitkgp:
@yeban

`graphit` essentially does most of the things internally. The only options exposed are ` $queryDiv, $graphDiv, index, howMany`. When I was writing the API, the original idea that came to my mind was to let users decide where they want to start drawing from and how many tracks they want to draw (kind of start to end co-ordinates).

Later when we implemented buttons, however, I used `index` more as a parameter to initialize a couple of elements for the first time. The behavior of `howMany` remained the same with a default value of 20.

The API can certainly be improved but I'm not sure what behavior do you expect from internally handling the *number of hits shown*.

IsmailM:
@yeban 

I have made the requested changes...

I couldn't reproduce the error that you spoke of so can you please check for me (if this doesn't work can you try this [older version](https://github.com/IsmailM/sequenceserver/tree/SVG_responsiveness) )...

Issue #145, opened 2015-02-26
Fails if BLAST+ databases corrupt or incorrect permission on them


malcook:
This may be true of other system calls as well, I have not checked.

c.f. https://groups.google.com/d/msgid/sequenceserver/e45e1000-18ea-41d3-92ff-3d16bfda085a%40googlegroups.com.

yannickwurm:
+1 
Thanks @malcook

yannickwurm:
@vivekiitkgp can you catch this?

vivekiitkgp:
@yannickwurm Sure. I'll have a look.

yannickwurm:
Actually this is two parts: 
    1 the specific error.   
    2. Seqserver should have never died in such an ugly manner but should have had a global error catching...

yeban:
I will take care of this @vivekiitkgp.

@malcook,
SequenceServer checks return value of blastdbcmd and everywhere required to the best of my knowledge. It's blastdbcmd that doesn't always exit with non-zero code even on failure. To work around that SequenceServer even checks for error patterns in stderr output. And blastdbcmd isn't consistent even there. So it's just this particular bug that needs to be addressed. Additionally, the error you faced might as well have been due to incorrect permission on some BLAST+ database files instead of corrupt databases. We have discussed this on the mailing list in past and are introducing a check in SequenceServer via #139.

malcook:
Hi - I should have been more specific as to line number.  I just now did a "gem install --pre sequenceserver" in order to see effect of your changes and found that still database.rb line #102  doe not check the return value of blastdbcmd.  This was the line that caused me to report this issue in the first place.
```
        cmd = "blastdbcmd -recursive -list #{database_dir}" \
              ' -list_outfmt "%f	%t	%p	%n	%l	%d" 2>&1'
        list = `#{cmd}`
        list.each_line do |line|
          name = line.split('	')[0]
          next if multipart_database_name?(name)
          self << Database.new(*line.split('	'))
        end
```

my problem was that one of my databases was in some sense corrupt.  I tested and blastdbcmd return code was non-zero in those cases.  

My suggestion is to filter out databases returning non-zero, report them with WARNINGS, and proceed.

yeban:
[...]

PR #191, opened 2015-05-13
Changes

alekhyam94:
1) Including length in tabular form.
2) Sorting by score when evalues are same.

### DIFF
--- a/lib/sequenceserver/blast/query.rb
+++ b/lib/sequenceserver/blast/query.rb
@@ -17,7 +17,8 @@ module SequenceServer
       end
 
       def sort_hits_by_evalue!
-        @hits = hits.sort_by(&:evalue)
+        # change made here
+        @hits = hits.sort_by{|h| [h.evalue, h.score]}
       end
 
       attr_reader :id, :title
--- a/views/result.erb
+++ b/views/result.erb
@@ -40,6 +40,7 @@
               <thead>
                 <th class="text-left"> Number </th>
                 <th>Sequences producing significant alignments</th>
+                <th class="text-right"> Length </th>
                 <th class="text-right"> Total score </th>
                 <th class="text-right"> E value </th>
               </thead>
@@ -50,6 +51,7 @@
                   <td> <a href="<%="#Query_#{query.number}_hit_#{hit.number}"%>"><%= "#{hit.id}" %></a> </td>
                   <td class="text-right"><%= prettify hit.score  %></td>
                   <td class="text-right"><%= prettify hit.evalue %></td>
+                  <td class="text-right"><%= prettify hit.length %></td>
                 </tr>
               <% end %>
               </tbody>


lib/sequenceserver/blast/query.rb
```diff
@@ -17,7 +17,8 @@ def initialize(*args)
       end
 
       def sort_hits_by_evalue!
-        @hits = hits.sort_by(&:evalue)
+        # change made here
+        @hits = hits.sort_by{|h| [h.evalue,h.score]}
```
vivekiitkgp: Space after `,`.

vivekiitkgp:
@yeban How important do we find the `hit.length` value? I'd like to discuss it a bit before merging.

yeban:
@alekhyam94 I would prefer the length column be placed after score and evalue. Also, need to put a space after `,` as @vivekiitkgp pointed out.

@vivekiitkgp I would have benefitted from this just last week when I was analysing the results of annotation liftover. And as I have mentioned in #145 it's helpful to quickly get an overview of length of the hits for manual curation.

vivekiitkgp:
While there is indeed no harm in doing this, these numbers are often quite large for a decent database and might occupy too much space (not a value for space trade).

Also, when we consider the immediate use case of Sequenceserver, if I'm blasting a query against a number of databases, the most important thing I'm concerned is with the quality of hits (it doesn't really matter from how big database do they come from, at least, during the start). I can always navigate to the respective hit and examine the database size if needed.

This is what I had in mind.

yeban:
Because you say "big database" I'm wondering if you are misunderstanding the issue. We are talking hit length here, not database size. Again, what's important is what the hits are saying about my query sequence. I think when blasting proteins and transcripts, length of the hit is almost always an important metric. For instance, if BLAST a predicted protein sequence against Uniprot and I see that all the strong hits are longer than my query sequence, it could mean the predicted protein is incorrect. And that is a significant use case of BLAST & SequenceServer.

lib/sequenceserver/blast/query.rb
```diff
@@ -17,7 +17,8 @@ def initialize(*args)
       end
 
       def sort_hits_by_evalue!
-        @hits = hits.sort_by(&:evalue)
+        # change made here
+        @hits = hits.sort_by{|h| [h.evalue, h.score]}
```
yeban: Very nicely done.

yeban:
Thanks @alekhyam94. Merged. You forgot to change table headers when you moved around the length column. I took care of that (ac6dea5302c8a7f622d74c1a53b9f9c5df10d453).

vivekiitkgp:
> Because you say "big database" I'm wondering if you are misunderstanding the issue.

Umm... No. I'm not. Of course `big database` was a typo but just blast that `multi_hsp` query against the `Si_gnF` database and you'll see that a lot of hits have `hit.length` in the range of 2 * 10 ^ 6. Clearly, that adds too little information.

> if BLAST a predicted protein sequence against Uniprot and I see that all the strong hits are longer than my query sequence, it could mean the predicted protein is incorrect.

I don't think that's a correct inference. In my idea, that doesn't mean anything significant at all. For example, in the previous quote the hit length is in million range but still there might be an exact match to my query.

Correct me if I'm wrong.

yeban:
Protein sequences are much smaller. And my example considers protein-protein BLAST(p).

vivekiitkgp:
Alright. Does no harm if you insist.

PR #206, opened 2015-09-17
Re-enabled testing

terfn:
Testing is now re-enabled, and all blast tests pass. Also addressing #196.

### DIFF
--- a/spec/sample_reports/blastn_sample/job.yaml
+++ b/spec/sample_reports/blastn_sample/job.yaml
@@ -1,0 +1,8 @@
+--- !ruby/object:SequenceServer::BLAST::Job
+id: blastn_sample
+method: blastn
+
+databases:
+- !ruby/struct:SequenceServer::Database
+  name: "../database/sample/transcripts/Solenopsis_invicta/Sinvicta2-2-3.cdna.subset.fasta"
+  type: nucleotide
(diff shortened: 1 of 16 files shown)


spec/sample_reports/blastn_sample/job.yaml
```diff
@@ -0,0 +1,8 @@
+--- !ruby/object:SequenceServer::BLAST::Job
+id: blastn_sample
+method: blastn
+
+databases:
+- !ruby/struct:SequenceServer::Database
+  name: "../database/sample/transcripts/Solenopsis_invicta/Sinvicta2-2-3.cdna.subset.fasta"
```
yeban: Did you add this path yourself? I think SS uses absolute path.

spec/sample_reports/blastn_sample/rfile
```diff
@@ -33,7 +33,7 @@ A79AC4AAD060E0B6108A80C5808E76AF0888D09A23F480E7630350B60D6863D2403C822D821E65
           }
         }
       },
-      subject database "spec/database/sample/transcripts/Solenopsis_invicta/Si
+      subject database "../database/sample/transcripts/Solenopsis_invicta/Si
```
yeban: If absolute path is given to blast, this will contain absolute path as well. Wondering if this together with prev comment might solve travis issue?

spec/sample_reports/blastn_sample/job.yaml
yeban: Even if it doesn't make or break anything, it should be absolute path so as to be representative of how SS works.

terfn: I just took this from the asn file, in job it was absolute. In the asn it was like this except I had to replace spec with ..

Issue #199, opened 2015-08-27
Add query coverage in the default output reports


billzt:
"Query coverage" is one of the shown items in the NCBI public BLAST server, and it could also be automatically calculated by BLAST+ command line tools (in -outfmt options). What's more, it is an important score to judge homology. Therefore, it is very necessary to add this items in the default output page.

yeban:
The reason query coverage is currently not included in the HTML report is because coverage info is not present in BLAST's XML output, which we use to generate the HTML report. To get coverage (and taxonomy) info SequenceServer will additionally have to generate and parse tabular output. When this is implemented:

1. Query coverage (`qcov`) should be included in the tabular summary of hits per query.
2. Additionally, per hsp coverage (`qcovhsp`) should be included in HSP stats.

billzt:
@yeban Good idea!

PR #211, opened 2015-09-20
Query Coverage & Sciname

terfn:
Addressing #199. Also added scientific name to tsv report.


lib/sequenceserver/blast/hit.rb
```diff
@@ -3,13 +3,14 @@ module SequenceServer
   module BLAST
     # Hit Object to store all the hits per Query.
     Hit = Struct.new(:query, :number, :id, :accession, :title,
-                     :length, :hsps) do
+                     :length, :qcovs, :sciname, :hsps) do
```
yeban: Is it qcovs or qcov?

terfn: qcovs everywhere

PR #212, opened 2015-09-20
Add explicit download path message

vivekiitkgp:
If a the download was complete but setup failed, user might be in fix when asked again for download during next attempt. Clearly state the download path.

### DIFF
--- a/bin/sequenceserver
+++ b/bin/sequenceserver
@@ -8,8 +8,8 @@ ENV['RACK_ENV'] ||= 'production'
 # display name for tools like `ps`
 $PROGRAM_NAME = 'sequenceserver'
 
-# Given a url downloads the archive into a temporary file
-# and and extratcs it into ~/.sequenceserver
+# Given a url, downloads the archive into a temporary file
+# and and extracts it into ~/.sequenceserver.
 def download_from_url(url)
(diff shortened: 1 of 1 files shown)


vivekiitkgp:
An alternative would be to save the config file after each step.

bin/sequenceserver
```diff
@@ -245,6 +245,8 @@ MSG
           puts
           puts 'Installing NCBI BLAST+.'
           puts "RUBY_PLATFORM #{RUBY_PLATFORM}"
+          puts 'Archive will be extracted in ~/.sequenceserver directory.'
```
yeban: @vivekiitkgp How about we instead change the 'installing ncbi blast+' msg (L246) to 'Installing NCBI BLAST+ to ~/.sequenceserver/ncbi-...'?

vivekiitkgp: @yeban It should work as well. Although, I'm not sure how to interpret it if the *installation* fails. Was it installed? We can be even more verbose I guess.

yeban: There's ambiguity in 'Archive will be extracted ...' msg too. Nonetheless, this definitely improves on what we have, so I'm merging the PR.

PR #215, opened 2015-09-23
Fixing double e value

terfn:

This addresses the problem that when evalue is specified it’s once shown as expect and once as e value. This fix is to forget the user-supplied e value, since it is already part of the blast xml output, and rename the expect key to e value.

### DIFF
--- a/lib/sequenceserver/blast/report.rb
+++ b/lib/sequenceserver/blast/report.rb
@@ -85,9 +85,10 @@ module SequenceServer
       # matrix, evalue, gapopen, gapextend, and filters are available from XML
       # output.
       def extract_params(ir)
-        @params = Hash[
+        params = Hash[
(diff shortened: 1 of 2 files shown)


lib/sequenceserver/blast/report.rb
```diff
@@ -85,9 +85,10 @@ def extract_program_info(ir)
       # matrix, evalue, gapopen, gapextend, and filters are available from XML
       # output.
       def extract_params(ir)
-        @params = Hash[
+        params = Hash[
           *ir[7].first.map { |k, v| [k.gsub('Parameters_', ''), v] }.flatten
         ].merge(job.advanced_params)
+        @params = Hash['e value', params.delete('expect')].merge!(params)
```
terfn: This will ensure that 'e value' is at the front of the hash, so it will be the first column in the table at the top of the report

Issue #222, opened 2015-11-05
Travis failing


yannickwurm:
its ok for development version (master) to fail, but seeing a red "failing" button here or on biogems can be discouraging to potential users. 

Thus biogems should indicate travis status of  latest release branch; the readme could include both (clearly separating what is stable from dev)...

yeban:
Just realised that #223 (and follow-up) is not going to auto-resolve badge issue in README or on biogems.info.

/cc @vivekiitkgp

yeban:
The branch will have to be specified in the badge's URL. Easy to do in the README. Dunno how to do it for biogems.

yeban:
Fixed in README. Discusssing this on biogems: https://github.com/pjotrp/biogems.info/issues/83

yeban:
It's done.

PR #223, opened 2015-11-09
Update travis build, use latest release

vivekiitkgp:
#222 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,16 +3,27 @@ rvm:
   - 1.9.3
   - 2.0.0
   - 2.1.5
-  - 2.2.0
+  - 2.2.3
 
 addons:
+  apt:
+    packages:
+    - libqtwebkit-dev
+    - xvfb
+
   code_climate:
     repo_token:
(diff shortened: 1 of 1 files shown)


yeban:
Thanks!

yeban:
Ok. This has a few problems (apart from the one I fixed - fa73005940b1d75866ad5b3cec40ae5d0aeb03a6). It builds 1.0.4, not 1.0.5 which is the latest. This is because I forked off the new release into 1.0.x branch so that dev for 1.1 can continue on master. @vivekiitkgp would you be able to address this in travis.yml?

vivekiitkgp:
@yeban I didn't prefer checking out to a particular release because it is dependent on version numbering. Later if we make a 1.2.x release, it will require a change. Although, it is not a big deal.

Also, I had noticed the current release being shown as 1.0.4 instead of 1.0.5. It can be easily fixed by drafting a new release, I presume.

yeban:
Ok. After some trial and error I have settled on this - https://github.com/wurmlab/sequenceserver/blob/master/.travis.yml. Applying the updated Travis config to 1.0.x branch has it passing on Travis - hopefully the badge will update soon (both in README and on biogems)

We will have to update Travis config when master is ready. But that's okay. And in future (for 1.2.x etc.), this scenario can hopefully be avoided.

PR #224, opened 2015-11-13
[Revise request] Add the global coordinate system for matched regions

hkmoon:
Hi Anurag,

I added some codes for handling the global coordinate system for matched regions. Surely, there must be bugs in it. However, you can have a look of it if it's useful or not. Actually, this method is used in our PlanMine database for checking the transcripts. It would be nice to have an option to select the coordinate systems and color lookup tables.

Thanks a lot,
HongKee

sequenceserver__custom_blast_server

### DIFF
--- a/public/css/custom.css
+++ b/public/css/custom.css
@@ -350,7 +350,7 @@ svg:not(:root) {
 .axis path,
 .axis line {
     fill: none;
-    stroke: #f00;
+    stroke: #777;
     stroke-width: 2px;
     shape-rendering: crispEdges;
 }
--- a/public/js/jquery.t.js
+++ b/public/js/jquery.t.js
@@ -336,11 +336,11 @@
             var defs = svg.append('defs');
 
             var markerData = [
-                { id: 0, name: 'Arrow0', path: 'M 0,3 v-3 l6,3 l-6,3 Z', viewbox: '0 0 6 6', fill: '#de0007' },
-                { id: 1, name: 'Arrow1', path: 'M 0,3 v-3 l6,3 l-6,3 Z', viewbox: '0 0 6 6', fill: '#de00bf' },
-                { id: 2, name: 'Arrow2', path: 'M 0,3 v-3 l6,3 l-6,3 Z', viewbox: '0 0 6 6', fill: '#00c200' },
-                { id: 3, name: 'Arrow3', path: 'M 0,3 v-3 l6,3 l-6,3 Z', viewbox: '0 0 6 6', fill: '#00bebb' },
-                { id: 4, name: 'Arrow4', path: 'M 0,3 v-3 l6,3 l-6,3 Z', viewbox: '0 0 6 6', fill: '#2415bf' }
+                { id: 0, name: 'Arrow0', path: 'M 1,3 l-1,-3 l6,3 l-6,3 Z', viewbox: '0 0 6 6', fill: '#de0007' },
+                { id: 1, name: 'Arrow1', path: 'M 1,3 l-1,-3 l6,3 l-6,3 Z', viewbox: '0 0 6 6', fill: '#de00bf' },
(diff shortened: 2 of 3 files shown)


yannickwurm:
THanks HongKee - this looks cool. I have some comments/questions: 
1. I think the arrowheads look ambiguous: they look like the alignment goes further than it actually does. How about just putting them at the end of the alignment?
2. what happens when the hit sequence is longer than the query? 
3. what happens if hit or query are repetitive and match multiple times
4. what happens when hit is palindromic? 
5. Have you had a look at how uniprot do their blast? they basically add an additional visualisation of the hit sequences (in a different column) to show which parts of it align to the query. Have you considered something alon gthose lines as well?

Kind regards, Yannick

hkmoon:
Dear Yannick, 

Thank you so much for your comments. Most of them have not yet been considered because I focused on only my cases. Please, feel free to share other use cases in order to cover all types of blast search result visualization.

> 1\. I think the arrowheads look ambiguous: they look like the alignment goes further than it actually does. How about just putting them at the end of the alignment? Alternatively, the aligning part could be filled with a pattern that looks similar to <<<< or >>>>> rather than just being a solid line

That is an actual brilliant idea. I will follow your advice.

> 2\. what happens when the hit sequence is longer than the query?

If the hit sequence is longer than query, the canvas tries to draw the whole sequence. Then, the query might look very tiny if the hit sequence is much longer. It makes a problem such a case. That's why I am considering a zooming function.

> 3\. what happens if hit or query are repetitive and match multiple times

Sorry that I never thought about this. I would like to test this case too.

> 4\. what happens when hit is palindromic?

This is not considered too. Can you provide some examples in order to test 3 and 4 cases?

> 5\. Have you had a look at how uniprot do their blast? they basically add an additional visualisation of the hit sequences (in a different column) to show which parts of it align to the query. Have you considered something alon gthose lines as well?

It is quite interesting to see it. It would be great to have this way of visualization in SequenceServer as well. However, I have not yet thought about it actually. Let's think more which is a better approach for the visualization in SequenceServer.

Thank you so much for the additional comments which I have not considered enough. I will get back to you as I improve the codes based on your advises. If you do not mind, could I ask some guides to improve the codes too? Actually, my background is not bio-informatics. So, it takes some time for me to understand your terminologies sometimes.

Best Regards,
HongKee

yeban:
Dear @hkmoon 

I am really sorry I have had the chance only now to go through your PR in detail. I am not sure I can accept it, for the following reasons:

1. The arrows look like that's where the alignment ends. I much prefer @yannickwurm's suggestion of filling the hsps with >> or << pattern.
2. Showing how much a hit extends beyond the matched regions with a line is great for RNA and proteins. But for genomic sequences you can end up with matched region having very low coverage, and the extending lines being to look weird. I am not sure Google Map or JBrowse like zooming is the most appropriate solution either. You may end up zooming and panning through large chunks of unmatched regions: there won't be much point to that.
3. Colour scheme - I am open to switching from grey gradient to blue, yellow, orange, red in order of decreasing significance (like the colours of a flame). I am sorry I don't like this colour scheme.

Further building on Yannick's feedback,
a. I guess overlapping hsps can be drawn on different lines, one below another in order of decreasing significance.
b. Palindromic hits could be centromeric regions or stretches of di- nucleotide repeats (ATATATAT). I don't think the alignments overview needs to do anything about such cases.

We recently concluded incorporating 3 new visualisations on the report page and making some improvements to alignments overview. Now would be the perfect time to re-think and improve it. I would be very interested in 1, 3, a.

hkmoon:
Dear @yeban 

I am also very sorry that I could not keep working based on @yannickwurm's suggestions because there were prioritized projects to be done. However, I am still interested in SequenceServer's visualization approach and discussions with you as well. I would be open to 1, 3, a. However, 2 is a special requirements of our institute client.

Thank you a lot for the detailed reasons for the PR rejection. While you manage such a big project like SequenceServer, you have to branch in/out wisely in order to succeed in the project. I will keep your suggestion in mind and might need to evolve my forked source according to our institute clients' (who are using Intermine integration) requirements. When I implement a good feature, I will keep you updated.

All the best,
HongKee

Issue #229, opened 2015-11-15
New Reports Page Bugs


terfilip:
report.js doesn't get properly loaded first time I run the page, and clicking on FASTA of all hits results in a request to /undefined. Then going back suddenly lets the report.js be loaded and it works fine.

Will add more bugs on the reports page here as I find them.

yeban:
From the top of my head:
- Kablammo needs to be responsive, like graphical overview.
- sequence viewer links work only after page load.

IsmailM:
Another new Reports Page bug - Sidebar links need to be disabled (or fully greyed out) until that BLAST results load fully. Clicking on the Sidebar links before the BLAST results are produced causes the following error:

```
Argument Error - Database id should be one of: cceba041efceb36c84cc5067938a0e89
```

yeban:
I think sidebar shouldn't be shown at all until the results are fully loaded.

PR #230, opened 2015-11-17
Removing 30 hit FASTA download limit

terfn:

Also fixes the first point of #229. 

If this is ok, I will add automatic removal of the yaml files along with removing the job folders in my next PR.

### DIFF
--- a/lib/sequenceserver/routes.rb
+++ b/lib/sequenceserver/routes.rb
@@ -98,14 +98,18 @@ module SequenceServer
       sequence_ids = params[:sequence_ids].split(/\s/)
       database_ids = params[:database_ids].split(/\s/)
 
-      sequences = Sequence::Retriever.new(sequence_ids, database_ids,
-                                          params[:download])
+      sequences = Sequence::Retriever.new(sequence_ids, database_ids)
+      sequences.to_json
+    end
+
+    post '/get_sequence' do
+      sequences = Sequence::Retriever.new(params["sequence_ids"].split(","),
+                                          params["database_ids"].split(","),
+                                          true)
 
       send_file(sequences.file.path,
                 :type     => sequences.mime,
-                :filename => sequences.filename) if params[:download]
-
-      sequences.to_json
+                :filename => sequences.filename)
     end
 
     # Download BLAST report in various formats.
--- a/public/js/report.js
+++ b/public/js/report.js
@@ -975,40 +975,47 @@ var SideBar = React.createClass({
     },
 
     /**
+     * Dynamically create form and submit.
+     */
+    postForm: function (database_ids, sequence_ids) {
+      var form = $("<form/>").attr('action', '/get_sequence').attr('method', 'post');
+
+      addField("sequence_ids", sequence_ids);
+      addField("database_ids", database_ids);
+
+      form.appendTo('body').submit().remove();
+
+      function addField(name, val) {
+        var input = $("<input>").attr("type", "hidden")
+                                .attr("name", name)
+                                .val(val);
+        form.append(input);
+      }
+    },
+
+    /**
      * Event-handler for downloading fasta of all hits.
      */
     downloadFastaOfAll: function () {
         var num_hits = $('.hitn').length;
 
         var $a = $('.download-fasta-of-all');
-        if (num_hits >= 1 && num_hits <= 30) {
+        if (num_hits > 0) {
             var sequence_ids = $('.hit-links :checkbox').map(function () {
                 return this.value;
             }).get();
-            $a
-            .enable()
-            .attr('href', this.generateURI(sequence_ids, $a.data().databases))
-            .tooltip({
+            this.postForm($a.data().databases, sequence_ids);
+            $a.tooltip({
                 title: num_hits + " hit(s)."
             });
-            return;
         }
-
-        if (num_hits === 0) {
-            $a.tooltip({
+        else {
+            $a
+              .disable()
+              .tooltip({
                 title: "No hit to download."
             });
         }
-
-        if (num_hits > 30) {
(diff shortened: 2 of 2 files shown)


yeban:
Thanks @terfn. This looks it will work. However, I don't understand your rationale for a few things:

1. Why differentiate b/w less than 30 and greater than 30 sequences? Why not fetch all via POST?
2. Are two new API endpoints required? Can't you initiate file download on POST?

For `routes.rb`, I had the following in mind. `get '/get_sequence'` serves sequence viewer, which fetches one sequences at a time and the return value is JSON. `post '/get_sequence'` would cater to downloads, one FASTA or bulk.

terfn:
@yeban ok all will be post. Switched to submitting form with jQuery, which fixes your second point

lib/sequenceserver/routes.rb
```diff
@@ -108,6 +108,17 @@ class Routes < Sinatra::Base
...
 
+    post '/get_sequence' do
+      sequences = Sequence::Retriever.new(params["sequence_ids"].split(","),
+                                          params["database_ids"],
+                                          params["download"])
+
+      send_file(sequences.file.path,
+                :type     => sequences.mime,
+                :filename => sequences.filename) if params["download"]
+      sequences.to_json
```
yeban: This line and the conditional in L118 is now redundant, right? You can simply pass `true` to `Sequence::Retriever` instead of `params['download']`.

 Similarly in `get '/get_sequence'` the `send_file` line, right?

terfn: @yeban ok removed. I just wan't sure if there wasn't some other functionality that used this.

lib/sequenceserver/routes.rb
```diff
@@ -99,13 +99,21 @@ class Routes < Sinatra::Base
       database_ids = params[:database_ids].split(/\s/)
 
       sequences = Sequence::Retriever.new(sequence_ids, database_ids,
-                                          params[:download])
+                                          true)
 
       send_file(sequences.file.path,
                 :type     => sequences.mime,
-                :filename => sequences.filename) if params[:download]
+                :filename => sequences.filename)
```
yeban: Nope. This one needs to send json.

lib/sequenceserver/routes.rb
```diff
@@ -99,13 +99,19 @@ class Routes < Sinatra::Base
       database_ids = params[:database_ids].split(/\s/)
 
       sequences = Sequence::Retriever.new(sequence_ids, database_ids,
-                                          params[:download])
+                                          true)
```
yeban: The third parameter (`true`/`false`) to `Sequence::Retriever.new` indicates whether the sequences retrieved from blast databases should be written to a file (`true`) or not (`false`). Default is `false`.

While passing `true` here will work. The code is mis-leading and does additional work that's not required.

lib/sequenceserver/routes.rb
```diff
@@ -99,13 +99,19 @@ class Routes < Sinatra::Base
...
+
+      sequences.to_json
+    end
+
+    post '/get_sequence' do
+      sequences = Sequence::Retriever.new(params["sequence_ids"].split(","),
+                                          params["database_ids"],
```
yeban: No splitting of database ids required? Have you tested it with multiple dbs?

public/js/report.js
```diff
@@ -975,40 +975,45 @@ var SideBar = React.createClass({
     },
 
     /**
+     * Dynamically create form and submit.
+     */
+    postForm: function ($a, sequence_ids) {
```
yeban: The only use of `$a` in the entire function is to get database ids. Why not pass database ids as an argument instead?

public/js/report.js
```diff
@@ -975,40 +975,45 @@ var SideBar = React.createClass({
...
+      var form = $("<form action=\"/get_sequence\" method=\"post\"></form>")
```
yeban: +1 for dynamic form creation - one function abstracts away the entire process. I had a static form in mind (generated through Report's render method) whose field would be updated, submitted, and cleared on click.

Consider breaking the string using other jquery methods. I don't know if all jquery methods are available to an unattached dom node created this way. But if `prop` is, that should do: `$('<form/>').prop('action', ...).prop('method', ...)`.

public/js/report.js
```diff
@@ -975,40 +975,45 @@ var SideBar = React.createClass({
...
+            $a.tooltip({
                 title: num_hits + " hit(s)."
             });
-            return;
         }
-
-        if (num_hits === 0) {
+        else {
             $a.tooltip({
                 title: "No hit to download."
```
yeban: The link should additionally remain disabled if nothing to download.

public/js/report.js
```diff
@@ -1021,36 +1026,23 @@ var SideBar = React.createClass({
...
                 title: num_checked + " hit(s) selected."
             });
             return;
         }
-
-        if (num_checked === 0) {
+        else {
             $n.empty();
             $a.tooltip({
                 title: "No hit selected."
```
yeban: Same as above - the link should remain disabled if nothing to download.

terfn:
@yeban updated

public/js/report.js
```diff
@@ -975,40 +975,48 @@ var SideBar = React.createClass({
     },
 
     /**
+     * Dynamically create form and submit.
+     */
+    postForm: function (database_ids, sequence_ids) {
+//var form = $("<form action=\"/get_sequence\" method=\"post\"></form>")
```
vivekiitkgp: Stray comment line. Please remove if it isn't need.

yannickwurm:
+1 from Sam Neuenschwander

PR #231, opened 2015-11-18
Automatically remove jobdirs

terfn:

Periodically removes job directories in dotdir. The user should be able to define the lifetime of each job dir. The default is 24 hours. The time can be specified in the config file using the format
“minutes-hours-days”(i.e. "08-10-12"). If the string “INF” is given, no files will ever be deleted. If only one number is specified, hours will be assumed.

### DIFF
--- a/lib/sequenceserver/job.rb
+++ b/lib/sequenceserver/job.rb
@@ -12,6 +12,7 @@ module SequenceServer
   # Sub-classes must at least define `run` instance method.
   class Job
     DOTDIR = File.expand_path('~/.sequenceserver')
+    UUID_PATTERN = /[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89aAbB][a-f0-9]{3}-[a-f0-9]{12}/
 
     class << self
       # Creates and queues a job. Returns created job object.
@@ -26,6 +27,15 @@ module SequenceServer
       # TODO: What if the given job id does not exist?
       def fetch(id)
         YAML.load_file(File.join(DOTDIR, id, 'job.yaml'))
+      end
+
+      def all
+        jobdirs = Dir["#{DOTDIR}/*"].select{ |f| f =~ UUID_PATTERN }
+        jobdirs.map{ |fname| fetch File.basename(fname) }
+      end
+
+      def delete(id)
+        FileUtils.rm_r File.join(DOTDIR, id)
       end
 
       private
@@ -57,7 +67,7 @@ module SequenceServer
       raise e
     end
 
-    attr_reader :id
+    attr_reader :id, :completed_at
 
     # How to execute the job.
     #
@@ -101,6 +111,7 @@ module SequenceServer
 
     # Marks the job as done.
     def done!
+      @completed_at = Time.now
       @done = true
       save
     end
--- a/lib/sequenceserver/job_remover.rb
+++ b/lib/sequenceserver/job_remover.rb
@@ -1,0 +1,65 @@
+require 'sequenceserver/job.rb'
+
+module SequenceServer 
+  class JobRemover
+
+    DEFAULT_LIFETIME = 86400
+
+    def initialize(lifetime)
+      parse lifetime
+      spawn_cleanup if @clean_needed
+    end
+
+    extend Forwardable
+    def_delegators SequenceServer, :logger
+
+    def spawn_cleanup
+
+      Thread.new do 
+        loop do
+          begin
+            #leave incomplete jobs
+            finished_jobs = Job.all.select{ |f| f.done? }
+            now = Time.now
+
+            unless finished_jobs.empty?
+              exp_jobs = finished_jobs.select do |job|
+                (job.completed_at + @lifetime) < now
+              end
+
+              exp_jobs.each{ |job| Job.delete job.id }
+            end
+
+            remaining_jobs = (exp_jobs.nil?) ? finished_jobs : finished_jobs - exp_jobs
+            oldest_time = remaining_jobs.map(&:completed_at).min
+
+            @next_cleanup = @lifetime
+            if  !!oldest_time && now < (oldest_time + @lifetime)
+              @next_cleanup = (oldest_time + @lifetime) - now
+            end
+
+            sleep(@next_cleanup.to_i)
+          rescue => e #StandardError
+            logger.debug("Cleanup thread shutting down due to Error:")
+            logger.debug(e.inspect)
+            logger.debug(e.backtrace.join("\n"))
+
+            Thread.exit
+          end
+        end
+      end
+    end
+
+    def parse(job_lifetime)
+      @clean_needed = true
+      if job_lifetime.nil?
+        @lifetime = DEFAULT_LIFETIME
+      elsif job_lifetime == "INF"
+        @clean_needed = false
+      else
+        @lifetime = job_lifetime.to_i * 60
+      end
+    end
+
+  end
+end
--- a/lib/sequenceserver.rb
+++ b/lib/sequenceserver.rb
@@ -9,6 +9,7 @@ require 'sequenceserver/sequence'
 require 'sequenceserver/database'
 require 'sequenceserver/blast'
 require 'sequenceserver/routes'
+require 'sequenceserver/job_remover'
 
 # Top level module / namespace.
 module SequenceServer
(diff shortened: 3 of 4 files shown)


lib/sequenceserver/config.rb
```diff
@@ -61,6 +62,26 @@ def normalize(data)
       data
     end
 
+    def parse_job_lifetime data
```
vivekiitkgp: Please add suitable tests for this as well.

terfn: @vivekiitkgp ty :)

yeban:
Thanks @terfn.

If I set job lifetime to 1 month and restart SequenceServer on 29th day, then month old jobs will only be deleted after another 30 days. That is, after 59 days from when the job was created. The thread should thus sleep for the difference of job lifetime and the oldest job that still exists.

If I set job lifetime to 1 hour and queue a BLAST job that takes more than 1 hour, the job will be deleted even before it has finished. Thus only jobs that are done should be deleted. I think we should include a `completed_at` key to the job and base calculations on that.

The syntax for defining job lifetime will be clear to those who have worked with cron. While flexible, I find cron's syntax a little confusing. I think we should only allow defining job lifetime in minutes.

I think the functionality should be captured in a separate class, in a separate file. Job lifetime should be passed to it, and it should decide whether to spawn a thread or not. It should work with an array of job objects returned from `Job.all` method.

terfn:
@yeban ok there is now a separate file, job lifetime can only be in minutes, and the lifetime is the number of minutes the job folder will live after the job has completed. Only completed jobs will ever be removed

lib/sequenceserver/blast/job.rb
```diff
@@ -36,7 +36,7 @@ def initialize(params)
       # Attributes used by us.
       #
       # Should be considered private.
-      attr_reader :method, :qfile, :databases, :options
+      attr_reader :method, :qfile, :databases, :options, :completed_at
```
yeban: I think `completed_at` should be defined in `Job` superclass.

lib/sequenceserver/blast/job.rb
```diff
@@ -48,6 +48,7 @@ def run
 
         system("#{command} > #{rfile} 2> #{efile}")
         done!
+        @completed_at = Time.now
```
yeban: Wouldn't it be better handled if `done!` takes care of this?

lib/sequenceserver/blast/job.rb
```diff
@@ -88,6 +89,10 @@ def efile
         File.join(dir, 'efile')
       end
 
+      def delete_from_dotdir
```
yeban: I think this should be a class method on `Job` (superclass), if at all required. It's similar to `create` and `fetch` methods defined on `Job`.

In (modern) OOP, it's often a good idea to define instances only as data objects and define methods that work on the data objects, "the services", in another module. In Ruby, class methods often work well to define a "service object", especially if there's only one kind of service around the data object - in this case create, read, update, delete (CRUD).

lib/sequenceserver/job_remover.rb
```diff
@@ -0,0 +1,57 @@
...
+    def spawn_cleanup
+      Thread.abort_on_exception = true if ENV['RACK_ENV'] == 'development'
+
+      Thread.new do 
+        loop do
+          #leave incomplete jobs
+          current_jobs = Job.all.select { |f| !f.completed_at.nil? }
+          oldest_time = Time.at(0)
+
+          unless current_jobs.length == 0
```
yeban: Prefer `Array#empty?`.

yeban:
Thanks @terfn. This already looks good. I will go through the main logic (`spawn_cleanup`) in afternoon. In the meantime, I have left some quick comments on the code. Let me know if you disagree with any of the points raised.

lib/sequenceserver/job_remover.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    DEFAULT_LIFETIME = 86400
+
+    def initialize(lifetime)
+      parse lifetime
+      spawn_cleanup if @clean_needed
+    end
+
+    def spawn_cleanup
+      Thread.abort_on_exception = true if ENV['RACK_ENV'] == 'development'
```
yeban: This a good measure. However, this changes the behaviour of all threads that are spawned by SequenceServer, including worker threads that run BLAST. Thus this should be configured during program initialisation.

Also, what happens if an exception is raised in this thread in production? The thread and thus job cleaner will terminate _silently_. I think it's better that you rescue all exceptions inside the loop, log them, and continue with the next tick.

lib/sequenceserver/job_remover.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      Thread.new do 
+        loop do
+          #leave incomplete jobs
+          current_jobs = Job.all.select{ |f| !f.completed_at.nil? }
```
yeban: This is leaky abstraction. Select by `Job#done?`. s/current_jobs/finished_jobs.

lib/sequenceserver/job_remover.rb
```diff
@@ -0,0 +1,59 @@
...
+          oldest_time = Time.at(0)
+
+          unless current_jobs.empty?
+            oldest_time = current_jobs.max do |left, right|
```
yeban: Why not `finished_jobs.map(&:completed_at).max`?

yeban: Shouldn't it be `min`?

lib/sequenceserver/job_remover.rb
```diff
@@ -0,0 +1,59 @@
...
+              left.completed_at <=> right.completed_at
+            end.completed_at
+ 
+            exp_jobs = current_jobs.select do |job|
```
yeban: Take this out of the `unless`. If `finished_jobs` is an empty Array, `expired_jobs` will automatically be an empty Array as well. Checking for nil at `L38-40` won't be required either. This is simpler to reason about.

lib/sequenceserver/job.rb
```diff
@@ -28,6 +29,15 @@ def fetch(id)
         YAML.load_file(File.join(DOTDIR, id, 'job.yaml'))
       end
 
+      def all
+        jobdirs = Dir["#{DOTDIR}/*"].select{ |f| f =~ UUID_PATTERN }
+        jobdirs.map{ |fname| fetch File.basename(fname) }
+      end
+
+      def delete_from_dotdir(id)
```
yeban: Let's just call it `delete`.

lib/sequenceserver/job_remover.rb
```diff
@@ -0,0 +1,59 @@
...
+              (job.completed_at + @lifetime) < Time.now
+            end
+          end
+
+          next_cleanup = @lifetime
```
yeban: Can you set this as an instance variable? That will make it possible to query job cleaner's state from outside, for debugging, instrumenting, etc.

lib/sequenceserver/job_remover.rb
terfn: Oh right, I was thinking of most recent time

lib/sequenceserver/job_remover.rb
```diff
@@ -0,0 +1,59 @@
...
+          unless current_jobs.empty?
+            oldest_time = current_jobs.max do |left, right|
+              left.completed_at <=> right.completed_at
+            end.completed_at
+ 
+            exp_jobs = current_jobs.select do |job|
+              (job.completed_at + @lifetime) < Time.now
+            end
+          end
+
```
yeban: We should log here how many jobs were deleted and when will the cleaner run next.

yeban:
`spawn_cleanup` has some logical inconsistencies:

1. `Time.now` is called several times. In particular, L26-28 means that even in the same sweep of the job cleaner, jobs are checked for expiration at different times. The current time should be captured once, at the top, and used as a reference throughout.
2. Oldest time is computed before deleting the jobs that have already been expired. Time till next sweep should be based on the oldest _remaining_ job.

Other than that, logging should be done to SequenceServer's standard logger. See `job.rb` and such to see how it's used.

I don't think `oldest_time` needs to default to epoch. If finished, unexpired jobs array is empty, min will return `nil` in which case you sleep for job lifetime, or compute how long to sleep.

yeban:
@terfn ping

terfn:
@yeban ok i think this might be better

Issue #188, opened 2015-04-16
Error when there's a * in hit sequence.


yeban:
Trying to reproduce the issue described at https://groups.google.com/forum/#!topic/sequenceserver/PGAPMssi5I4, I ran into another one.

```
  ArgumentError - invalid byte sequence in UTF-8:
/Users/priyam/sequenceserver/lib/sequenceserver/sequence.rb:197:in `split'
/Users/priyam/sequenceserver/lib/sequenceserver/sequence.rb:197:in `block in run'
/Users/priyam/sequenceserver/lib/sequenceserver/sequence.rb:196:in `each_line'
/Users/priyam/sequenceserver/lib/sequenceserver/sequence.rb:196:in `each'
/Users/priyam/sequenceserver/lib/sequenceserver/sequence.rb:196:in `map'
/Users/priyam/sequenceserver/lib/sequenceserver/sequence.rb:196:in `run'
/Users/priyam/sequenceserver/lib/sequenceserver/sequence.rb:172:in `initialize'
/Users/priyam/sequenceserver/lib/sequenceserver/routes.rb:127:in `new'
/Users/priyam/sequenceserver/lib/sequenceserver/routes.rb:127:in `block in <class:Routes>'
[...]
```

To reproduce:

```
$ mkdir /tmp/test
$ cd /tmp/test
$ echo '>gi|5835987|ref|NP_008750.1|ND6_15094 NADH dehydrogenase subunit 6 (mitochondrion) [Vidua chalybeata]
MMEFVLFLGLLFALGGLAVASNPSPYYGVVGLVVAAIAGGGWLVSLGVSFVSLMLVMVYLGGMLVVFVYS
VSLAADPYPETWTDWGVVGYGFGMGLVVMVGLVVSGVFGLLADVGTVNGAGLLSVRVDFSGAAVLYSEGA
GFLLIGGWGLLLTLFVVLELVRGLSRGAIRAV*' > prot.fa
$ makeblastdb -in prot.fa -dbtype prot -parse_seqids
$ sequenceserver -d .
```

Now BLAST with the same query sequence and try to view sequence or download FASTA.

pydupont:
I solved this bug by patching the run function in lib/sequenceserver/sequence.rb

```ruby
def run
  command = "blastdbcmd -outfmt '%g	%i	%a	%t	%s'" \
            " -db '#{database_names.join(' ')}'" \
            " -entry '#{sequence_ids.join(',')}'"
  
[...]

PR #239, opened 2016-02-01
Fix for issue #188

tomas-pluskal:
This pull request contains the proposed fix for issue #188


yeban:
See https://github.com/wurmlab/sequenceserver/issues/188#issuecomment-179768795.

PR #242, opened 2016-03-06
UI changes: Bigger font, remove text-justify, fix name inconsistency

vivekiitkgp:

cc @yeban 

Some of the changes I am not so very sure about. You may wish to review carefully.

IMO, text-justification is ugly. Only LaTeX does it right and we do not have that luxury here. We can still achieve much of it without making arbitrary spaces appear here and there using `letter-spacing`, `line-height` and `word-spacing`.

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -61,7 +61,7 @@
             </span>
           </button>
           <a
-            class="navbar-brand" href="/">
+            class="navbar-brand navbar-left" href="/">
             SequenceServer
             <small>
               1.0.7
@@ -108,7 +108,7 @@
             </li>
             <li>
               <a
-                href="https://github.com/yannickwurm/sequenceserver"
+                href="https://github.com/wurmlab/sequenceserver"
                 title="Will open GitHub in new window" target="_blank"
                 class="visible-lg">
                 <i class="fa fa-external-link" aria-hidden="true"></i>
--- a/css/seqserv.css
+++ b/css/seqserv.css
@@ -2,25 +2,33 @@
 body {
   font-family: "Source Sans Pro","Helvetica Neue",Helvetica,Arial,sans-serif;
   position: relative;
-  text-align: justify;
   padding-top: 70px;
+  letter-spacing: .02em;
+  word-spacing: .01em;
+  font-size: 16px !important;
 }
+
 section {
   padding-top: 50px;
   margin-top: -50px;
 }
+
 table, blockquote {
   font-size: inherit;
 }
+
 hr {
   margin-top: 0;
 }
+j
 code {
   color: inherit;
 }
+
 pre {
   border: none;
 }
+
 dl dd {
   padding-bottom: 10px;
 }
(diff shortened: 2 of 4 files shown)


yeban:
I find the website more readable now. But if previous version looked zoomed-out, this one looks zoomed-in. How to strike the right balance? Not sure about justifying vs left-aligning.

Can you fix http://vivekiitkgp.github.io/sequenceserver/ with a temporary commit so others can preview and provide feedback?

It's SequenceServer, not Sequenceserver. I can't consider the PR unless that's fixed.

vivekiitkgp:
> It's SequenceServer, not Sequenceserver.

It took the latter reference from our submitted paper. Does the paper needs
to be corrected or that's how things are supposed to be?

--
https://vivekiitkgp.github.io

vivekiitkgp:
@yeban Looks okay now. I will create a separate issue to incorporate Rumpt's feedback.

vivekiitkgp:
Okay. Should look okay now.

IsmailM:
The menu folds onto a new line on screens smaller than 1200px - personally I feel the menu looks better on a single line - what do you think? 

Secondly when the menu folds onto a second line, it covers the top 70px or so of the `.container` class.

Thus, I would suggest that for screen width of smaller than 1200px, the menu title font sizes should be decreased to 14px (so the menu doesn't fold) or a padding-top should be added to `.container` (so the top part of website can be seen)...

i.e. something like 

```css
/* small devices (i.e. 1200px and below) */
@media (max-width: 1200px) { 
   .navbar-nav>li>a {
     font-size:14px;
   }
}
```

or 

```css
/* small devices (i.e. 1200px and below) */
@media (max-width: 1200px) { 
   .container {
     padding-top: 68px;
   }
}
```

IsmailM:
Also, I feel the there shouldn't be a left border on the `.main` class when the sidebar is hidden. What do you think?

i.e. remove [this line](https://github.com/wurmlab/sequenceserver/blob/gh-pages/css/seqserv.css#L69) and then add the following:

```css
/* Medium devices (desktops, 992px and up) */
@media (min-width: 992px) { 
   .main {
     border-left: 1px solid #eee;
   }
}

```

Also any reason why the assets aren't minified - [Pagespeed Insights](https://developers.google.com/speed/pagespeed/insights/?url=http%3A%2F%2Fwww.sequenceserver.com%2F&tab=desktop) suggests the website can be improved quite a bit in terms of delivery of the assets - fixing those should be good for SEO.

yeban:
I will take care of these ... get back to you in evening.

yeban:
@IsmailM Take a look - http://sequenceserver.com/ (d0248dc0458eb9e520ffd76c247932dd0a9b2356).

vivekiitkgp:
@yeban Perhaps update the fontawesome (minified) as well?

yeban:
@IsmailM Assets are mostly minified. Looking at Pagespeed Insights, github images etc. can perhaps be server locally or maybe there's a different API call to get smaller image from github with proper cache headers set etc. Feel free to send a PR.

@vivekiitkgp Feel free to send a PR.

Issue #176, opened 2015-03-15
Sidebar Bug


IsmailM:
If you open SS in a browser with a width smaller than 992px (i.e. when the sidebar is hidden), the sidebar width is set as 0px and the top offset is set as 0px (set by some jQuery [here](https://github.com/yannickwurm/sequenceserver/blob/master/public/js/sequenceserver.js#L771)).

As a result, if you then increase the browser width from a width smaller than 992px, the sidebar looks really weird and does not work properly...

yeban:
Correct.

yannickwurm:
What is the easy solution to this?

yeban:
This was fixed sometime back.

PR #243, opened 2016-03-09
fix sidebar bug - #176

IsmailM:
Fixes Sidebar bug - #176

### DIFF
--- a/public/css/custom.css
+++ b/public/css/custom.css
@@ -487,3 +487,17 @@ svg:not(:root) {
   border-radius: 0;
   background-color: inherit;
 }
+
+/* Medium devices (desktops, 992px and up) */
+@media (min-width: 992px) { 
+  .sidebar.affix {
+    width: 243.5px;
+  }
+}
+
+/* Large devices (large desktops, 1200px and up) */
+@media (min-width: 1200px) {
+  .sidebar.affix {
+    width: 292.5px;
+  }
+}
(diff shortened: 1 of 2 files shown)


yeban:
We use fixed container width: 970 px at window width >= 992 px and 1170 px at window width >=1200 px. Sidebar takes 3 out of 12 columns in the grid: 25% of container width. Thus I would prefer setting sidebar width explicitly in CSS at 242.5 px and 292.5 px depending on screen size, using `@media` query. What do you think?

IsmailM:
Agreed - Since the Sidebar is within a `col-md-3`, the width is automatically set by Bootstap. And as such there is no need to add any CSS. Thus, I simply removed the JS line that set the sidebar width.

yeban:
Width should be set for `position: fixed` elements. And we thus should for `.sidebar`. If you remove the JS line that sets the width, you will see bottom borders in the sidebar shrink and grow as the sidebar is affixed and removed.

IsmailM:
The sidebar width is now set by CSS as requested.

yeban:
This is great. Please could you squash the commits? And don't build the assets - will do it during release.

IsmailM:
Done

yeban:
This doesn't seem to work: if sidebar was initially hidden, and then made to appear by resizing the window, the width is set right but the sidebar is no longer affixed.

yeban:
I patched that (1c78e31947e4ece90eb7250a00a6b04206eefca5).

PR #254, opened 2016-03-18
Fix the specs

audy:
Fixed a couple of specs.

### DIFF
--- a/lib/sequenceserver/blast/report.rb
+++ b/lib/sequenceserver/blast/report.rb
@@ -87,7 +87,8 @@ module SequenceServer
       def extract_params(ir)
         params = Hash[
           *ir[7].first.map { |k, v| [k.gsub('Parameters_', ''), v] }.flatten
-        ].merge(job.advanced_params)
+        ]
+        params = params.merge(job.advanced_params) unless job.advanced_params.nil?
         @params = Hash['e value', params.delete('expect')].merge!(params)
       end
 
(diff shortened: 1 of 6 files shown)


yeban:
I have re-enabled Travis. Try pushing a new or temporary commit maybe to trigger the build?

lib/sequenceserver/blast/report.rb
```diff
@@ -87,7 +87,8 @@ def extract_program_info(ir)
       def extract_params(ir)
         params = Hash[
           *ir[7].first.map { |k, v| [k.gsub('Parameters_', ''), v] }.flatten
-        ].merge(job.advanced_params)
+        ]
+        params = params.merge(job.advanced_params) unless job.advanced_params.nil?
```
yeban: I am curious if this change is precautionary or if nil is actually returned. If so, I think we should instead make sure job.advanced_params returns empty Hash instead of nil.

yeban:
Rest looks good to me.

lib/sequenceserver/blast/report.rb
audy: I was getting an error because `job.advanced_params` was `nil`.

audy: The reason why this is `nil` is that `job` is being loaded through a YAML file. It would be better to pass an actual instance of `Job` that way `parse_advanced` will get called and create the empty hash.

audy:
Specs are going to fail. I didn't fix all of them. But specs fail than a build of master. So you should still merge :smile_cat:

audy:
I think all of the remaining failing specs are related to not being able to find HTML elements with CSS selectors because React changed them.

PR #255, opened 2016-03-21
A few general improvements to website

vivekiitkgp:


IsmailM:
+1 for CDN, perhaps also use CDN for Jquery (Google's CDN) and Bootstrap JS (maxCDN)...

Secondly, I understand that the Bootstrap CSS file was generated using the [Bootstrap Customizer](https://github.com/wurmlab/sequenceserver/blob/gh-pages/css/bootstrap-3.3.6.min.css#L8) service and as such will be different to the version served under the CDN. Now, I'm not sure, how much the Bootstrap was customised, it may be better to simply to continue using the CDN version and add custom CSS where necessary.

IsmailM:
Also, how about splitting `Installation & configuration` into `Installation` and `Docs`, with `Docs` linking to the Documentation page?

_layouts/page.html
```diff
@@ -97,7 +97,7 @@
                 title="Will open in new window" target="_blank"
                 class="visible-lg">
                 <i class="fa fa-external-link" aria-hidden="true"></i>
-                Demo Server
+                Demo
```
yeban: Demo server

_layouts/page.html
```diff
@@ -106,7 +106,7 @@
                 title="Will open GitHub in new window" target="_blank"
                 class="visible-lg">
                 <i class="fa fa-external-link" aria-hidden="true"></i>
-                Source Code
+                Source
```
yeban: Source code

yeban:
How much impact does CDN make? I am not sure it will make too much of a difference. As far as I understand, GitHub is a CDN itself. Besides, as @IsmailM rightly pointed out, we use a customised version of Bootstrap CSS: we change the breakpoint at which top navbar will collapse. It's a one LESS variable change through BS customiser. Equivalent CSS likely will be more involved. The real bottleneck is the images. Please see the pagespeed insights report in #242 and following discussion.

IsmailM:
As Pagespeed showed, Github is not a CDN - it only caches for 10 minutes. And apparently Github Pages have no plans on fixing the HTTP Headers (see [here](http://stackoverflow.com/questions/14798589/github-pages-http-headers))

I would definitely suggest a CDN for the Jquery, Bootstrap JS, & FontAwesome. Firstly, they have a proper cache headers (unlike Github). Secondly, a large number of websites use these CDNs. Thus, there is a high probability that the asset has already been cached by the browser and thus the browser would not need to download another copy (of Jquery or bootstrap) to load the page - this means a faster page load.

Secondly, I’ll create a PR for addressing the rest of the points in the page speed insights - I’ve done most of the work (see [here](https://github.com/IsmailM/sequenceserver/tree/gh-pages), ignore the last commit) - I want to make a few further changes and tidy it up..

Lastly, while we’re talking about CDN, I would suggest that we use a CDN for everything - including images. This way we would have the right Headers for all assets.  One option would be to use the free plan in CloudFlare. Other options include [GitCDN](https://github.com/schme16/gitcdn.xyz) (which uses CloudFlare) or [RawGit](https://rawgit.com) (which uses MaxCDN).

yeban:
I cherry-picked b302ba9. Awaiting @IsmailM's PR re CDN (move to CDN completely or not at all and account for the customisation made to BS CSS). Don't like anchor tag's colour change.

yeban:
@IsmailM When can we see the PR re move to CDN?

PR #256, opened 2016-03-29
Website Update

IsmailM:
Currently work in progress...

**Done** 

- Lossless compression of local Images
- Get smaller user profiles avatars from Github (instead of getting a
larger image and then down-sizing)
- HTML compression
- Inline Fonts CSS (reduces the number of calls to the server)
- Use CDN for FontAwesome, Bootstrap JS, Jquery

**TODO**

- Use CDN for images
   - As discussed previously (#255), Github is not a CDN and as such does not cache any images - this isn’t a huge issue for images, but this could be solved by using CloudFlare’s free CDN - What are your thoughts on this, shall I get this done?

- CSS Optimisation - CSS should be merged into a single file (fewer calls to the server) & Critical CSS should be inlined (to avoid content blocking) 
    - In this case, I would suggest that we use [Uncss](https://github.com/giakki/uncss) to remove all the unused CSS (bootstrap CSS is massive). We could create a rake task to do this. 
     The file currently produced by Uncss is 4kb - this is much smaller than the current total of 157kb and it would be a single call to the server compared to 5 calls to the server...
   - Secondly, in order to ensure that the CSS isn’t content blocking, we could further extract the critical CSS (using something like [penthouse](https://github.com/pocketjoso/penthouse)) and then inline that into the HTML. Alternatively, since the Uncss results file is already quite small, we could just inline that (which is probably what I would suggest)…

- Lastly, I would suggest that we split the navbar menu tab `Installations & Configuration` into `Installation` and `Documentation`, with documentation linking to the documentation page.

<br><br>
@yeban , can you have a read of the above, and let me know of your thoughts on which (if any) of the TODO points I should implement...


vivekiitkgp:
Hi @IsmailM. Great work!

I think that the changes done so far are much better than we actually needed. The comments posted in TODOs list are great but aren't worth your effort in my opinion. I am not sure if we really have that many site hits in a day to require extensive optimization.

> I would suggest that we split the navbar 

Seems like a fair idea to me.

yannickwurm:
Agreed - there is absolutely no need for compression or load time optimization.

yeban:
I merged 4f22eb7 - cuts down the page load time by half.

I like the idea of eliminating extra css by using uncss and/or penthouse. Let's not inline CSS though. I have mixed feelings about minifying HTML - are you 100% sure it works? how much do we gain?

CDN - How do you upload stuff to CDN? Will there be a username/password and all? How will it impact rapid website modifications? Can it be fully streamlined with a rake task? If yes, then if you propose something mature I will accept it. But I don't really think it's required.

Making doc page stand out is important. If you split, Install and Config please do check by resizing browser window as well. Also, can you propose something to add some contrast and colors to the page?

PR #269, opened 2016-05-14
Added support for tmp files to be saved

aniarya82:
The current manner for downloading the xml or tsv files is to generate them and store them to temporary files and as soon as download of files are over, temporary files gets deleted. Now if user wants to again download them it will again be generated. To save the time consumed in it, why not save the files once generated. Here the files are being saved in the same folder as the archive_file, which is passed along for generating the xml and tsv files.

### DIFF
--- a/lib/sequenceserver/blast/formatter.rb
+++ b/lib/sequenceserver/blast/formatter.rb
@@ -24,7 +24,8 @@ module SequenceServer
       attr_reader :format, :mime, :specifiers
 
       def file
-        @file ||= Tempfile.new filename
+        # @file ||= Tempfile.new filename
+        @file = File.join(File.dirname(archive_file), filename)
       end
 
       def filename
@@ -38,11 +39,15 @@ module SequenceServer
         command =
           "blast_formatter -archive '#{archive_file}'" \
           " -outfmt '#{format} #{specifiers}'" \
-          " -out '#{file.path}' 2> /dev/null"
-        logger.debug("Executing: #{command}")
-        Dir.chdir(File.exist?(DOTDIR) && DOTDIR || Dir.pwd) do
-          system(command)
-        end
+          " -out '#{file}' 2> /dev/null"
+          if File.exist?(file)
+            logger.debug("File exist: #{file}")
+          else
+            logger.debug("Executing: #{command}")
+            Dir.chdir(File.exist?(DOTDIR) && DOTDIR || Dir.pwd) do
+              system(command)
+            end
+          end
       end
 
       def validate
--- a/lib/sequenceserver/blast/formatter.rb
+++ b/lib/sequenceserver/blast/formatter.rb
@@ -24,7 +24,6 @@ module SequenceServer
       attr_reader :format, :mime, :specifiers
 
       def file
-        # @file ||= Tempfile.new filename
         @file = File.join(File.dirname(archive_file), filename)
       end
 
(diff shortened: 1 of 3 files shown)


yannickwurm:
thanks - can we display a human-friendly error once the filesystem is full?

lib/sequenceserver/blast/formatter.rb
```diff
@@ -24,7 +24,8 @@ def initialize(archive_file, type)
       attr_reader :format, :mime, :specifiers
 
       def file
-        @file ||= Tempfile.new filename
+        # @file ||= Tempfile.new filename
```
yeban: Git tracks changes - commenting out what changed is not required.

lib/sequenceserver/blast/formatter.rb
```diff
@@ -38,11 +39,15 @@ def run
...
           "blast_formatter -archive '#{archive_file}'" \
           " -outfmt '#{format} #{specifiers}'" \
-          " -out '#{file.path}' 2> /dev/null"
-        logger.debug("Executing: #{command}")
-        Dir.chdir(File.exist?(DOTDIR) && DOTDIR || Dir.pwd) do
-          system(command)
-        end
+          " -out '#{file}' 2> /dev/null"
+          if File.exist?(file)
+            logger.debug("File exist: #{file}")
```
yeban: File exist**s**

lib/sequenceserver/blast/formatter.rb
```diff
@@ -38,11 +39,15 @@ def run
         command =
           "blast_formatter -archive '#{archive_file}'" \
           " -outfmt '#{format} #{specifiers}'" \
-          " -out '#{file.path}' 2> /dev/null"
-        logger.debug("Executing: #{command}")
-        Dir.chdir(File.exist?(DOTDIR) && DOTDIR || Dir.pwd) do
-          system(command)
-        end
+          " -out '#{file}' 2> /dev/null"
+          if File.exist?(file)
```
yeban: I would have placed the check the first thing in run method, logged and returned if file existed, and left the rest untouched.

lib/sequenceserver/blast/report.rb
```diff
@@ -44,8 +44,9 @@ def to_json
       # Generate report.
       def generate
         xml = Formatter.run(job.rfile, 'xml').file
-        tsv = parse_tsv Formatter.run(job.rfile, '__ssparse').file
-        ir = node_to_array(Ox.parse(xml.open.read).root)
+        tsv = parse_tsv File.read(Formatter.run(job.rfile, '__ssparse').file)
+        # ir = node_to_array(Ox.parse(xml.open.read).root)
```
yeban: Git knows what changed.

yeban:
This is a great start @aniarya82. Can you make the suggested changes?

aniarya82:
@yeban Thanks for the comments. I'll keep these in mind for future.

lib/sequenceserver/blast/formatter.rb
```diff
@@ -35,14 +35,18 @@ def filename
...
-          "blast_formatter -archive '#{archive_file}'" \
-          " -outfmt '#{format} #{specifiers}'" \
-          " -out '#{file.path}' 2> /dev/null"
-        logger.debug("Executing: #{command}")
-        Dir.chdir(File.exist?(DOTDIR) && DOTDIR || Dir.pwd) do
-          system(command)
-        end
+          if File.exist?(file)
+            logger.debug("File exist: #{file}")
+          else
```
yeban: Remove `else`. Return from the `if` block instead.

aniarya82:
@yannickwurm That is an great idea. I will look into it.

PR #270, opened 2016-05-15
pool.schedule method rename to pool.queue

aniarya82:
The pool.schedule method was a little misleading. We thought that the job is being scheduled to run at some specific time, but this isn't what happens. What happens is the job is queued to run as soon as server finishes up with previous jobs. So the pool.queue method instead of pool.schedule method makes more sense.

### DIFF
--- a/lib/sequenceserver/job.rb
+++ b/lib/sequenceserver/job.rb
@@ -42,7 +42,7 @@ module SequenceServer
 
       # Queues given job on the thread pool. Returns job.
       def queue(job)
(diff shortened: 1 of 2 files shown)


yeban:
First line of the commit message should be a small one line description of the change followed by a blank line and then a detailed summary or rational behind the change. e.g.,

    Rename Pool#schedule to Pool#queue.
    
    bla bla

aniarya82:
@yeban Thanks for the comment, will keep it in mind.

PR #274, opened 2016-05-24
error class added to handle all exceptions

aniarya82:
Initially all the exception was handled by Job, which sometime lead to unknown
error messages. This made user receive a message unrelated to what actually
issue was, which was misleading for the users. With a dedicated Error class we
can accurately set checks to catch correct issue and therefore provide an
appropriate error message to the user.


yannickwurm:
excellent

Issue #194, opened 2015-06-03
Ox::ParseError - invalid format, elements overlap at line ..., column ... [parse.c:561]


yeban:
From - https://groups.google.com/forum/#!topic/sequenceserver/QbE11EMQoEo

I can reproduce the error with Chris' XML file, and the error does look to be what Chris reported. Malformed XML: `<Iteration>` and `<Iteration_hits>` elements do not have a closing tag. Things seem to work if I manually add the corresponding closing tags.

The XML is generated malformed probably due to the following error raised during archive file to XML conversion with `blast_formatter`:

```
Error: Failed s_BlastXMLAddIteration Q(0/1NCBI C++ Exception:
    T0 "/home/coremake/release_build/build/PrepareRelease_Linux64-Centos_JSID_01_10349_130.14.22.10_9008__PrepareRelease_Linux64-Centos_1414443165/c++/compilers/unix/../../src/objtools/alnmgr/alnvec.cpp", line 902: Error: ncbi::objects::CAlnVec::TranslateNAToAA() - CAlnVec::TranslateNAToAA(): NA size expected to be divisible by 3
```

A gist of the XML and how to reproduce the issue - https://gist.github.com/yeban/b6a2c8fd082acc3e1371

Discussing if Ox can be lenient here - https://github.com/ohler55/ox/issues/111.

yeban:
@wwood What do you think? Is this a bug in BLAST or something wrong with Chris' and Tim's FASTA files?

yeban:
SequenceServer doesn't check if `blast_formatter` ran successfully. It assumes that once archive files has been successfully generated, conversion from archive to xml or tsv will not be problematic. Clearly, we will have to do error checking in the conversion step as well and surface appropriate error message to the user.

wwood:
Ho @yeban I don't have much to add - my suspicion is that it is a bug in blast, but need to reproduce on the cmdline in order to usefully report

yeban:
Pulling @tomas-pluskal into this thread who last reported the issue here - https://groups.google.com/d/msg/sequenceserver/72fE1lE_J7k/DlBJwc3VCAAJ

tomas-pluskal:
Hi.

I just installed 1.1.0-beta10 and created two files:

**rand1.fasta:**
````
>random_sequence
[...]

PR #275, opened 2016-05-24
Error handling

aniarya82:


yeban:
Thanks. On a quick look - the second commit doesn't entirely address #194 (which we discussed we want to address).

lib/sequenceserver/blast/formatter.rb
```diff
@@ -27,6 +27,10 @@ def file
         @file = File.join(File.dirname(archive_file), filename)
       end
 
+      def tmp
```
yeban: I think you should append to Job's efile.

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,76 @@
...
+        @job = job
+        callError
+      end
+
+      attr_reader :job
+
+      def callError
+        case job.exitStatus
+        when 1 # error in query sequence or options; see [1]
+          validate_options job.advanced_params # if advanced_params are incorrect
```
yeban: I don't understand why advanced params is being validated here and why is it done twice (once when a job is created and here - after job is done).

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,76 @@
...
+        File.size(efile) > 0
+      end
+
+      def readError
+        File.read(efile)
+        # IO.readlines(efile).last(3).each { |x| puts x } # for the -foo bar error.
+      end
+
+      def empty_rfile_error
+        rfile = File.join(DOTDIR, jid, 'efile')
```
yeban: Why not call `Job#rfile` directly?

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,76 @@
...
+        fail ArgumentError, "BLAST process was terminated prematurely." if File.zero?(rfile)
```
yeban: How do you conclude that empty rfile is ArgumentError? What is ArgumenError?

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,76 @@
...
+        true
+      end
+
+      def efile
```
yeban: What is the need for a proxy method?

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      attr_reader :job
+
+      def callError
+        case job.exitStatus
+        when 1 # error in query sequence or options; see [1]
+          validate_options job.advanced_params # if advanced_params are incorrect
+          empty_rfile_error # if rfile is empty
+        when 2, 3, 4, 255
+          fail ArgumentError, readError
```
yeban: How do you conclude this is ArgumentError? Especially in the light that original code called this scenario RuntimeError?

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      def allowed_chars
+        /\A[a-z0-9\-_\. ']*\Z/i
+      end
+
+      def disallowed_options
+        /-out|-html|-outfmt|-db|-query|-num_threads/i
+      end
+
+      def checkError
```
yeban: I don't see this function being called anywhere.

lib/sequenceserver/job.rb
```diff
@@ -88,6 +88,10 @@ def success?
       !!@success
     end
 
+    def exitStatus
```
yeban: When did we start using camel case?

lib/sequenceserver/blast/job.rb
```diff
@@ -51,32 +52,10 @@ def run
...
-        when 2, 3, 4, 255 # see [1]
-          efile.open
-          error = efile.read
-          efile.close
-          fail RuntimeError.new(status, error)
-        end
-
-        success!
+        status!(status)
+        success! if status == 0 && !File.zero?(rfile)
```
yeban: Also check if file exists. While I have never seen IO redirection not create a file, I think it's good to add it for sanity, completeness, and self-documentation.

lib/sequenceserver/job.rb
```diff
@@ -122,6 +126,11 @@ def success!
       save
     end
 
+    def status!(status)
```
yeban: Instead of creating a new method, I think we should modify `done!` to accept and save exit status.

lib/sequenceserver/blast/job.rb
```diff
@@ -51,32 +52,10 @@ def run
...
-        when 2, 3, 4, 255 # see [1]
-          efile.open
-          error = efile.read
-          efile.close
-          fail RuntimeError.new(status, error)
-        end
-
-        success!
+        status!(status)
+        success! if status == 0 && !File.zero?(rfile) && File.exist?(rfile)
```
yeban: Should check if file exists before testing it's empty or not, right?

lib/sequenceserver/blast/formatter.rb
```diff
@@ -39,10 +43,12 @@ def run
         command =
           "blast_formatter -archive '#{archive_file}'" \
           " -outfmt '#{format} #{specifiers}'" \
-          " -out '#{file}' 2> /dev/null"
+          " -out '#{file}' 2> #{tmp.path}"
         logger.debug("Executing: #{command}")
         Dir.chdir(File.exist?(DOTDIR) && DOTDIR || Dir.pwd) do
           system(command)
+          status = $CHILD_STATUS.exitstatus
+          fail ArgumentError, "Error in report generation" if status != 0
```
yeban: Need a more informative error message. I would show the stderr to user. And since this is for #194, I would additionally add "Maybe it's because sequence ids are not unique across database. See - https://groups.google.com/d/msg/sequenceserver/QbE11EMQoEo/nAcCOmlsSjwJ"

Also, did you verify if checking for exit status is enough to catch the issue? Going through mailing list archives indicates blast_formatter returns 0 in this case. And thus we must check if there's something on stderr or what.

lib/sequenceserver/blast/formatter.rb
yeban: Call it efile? Better name than tmp if not best. And consistent with Job.

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,43 @@
...
+        @job = job
+        call_error
+      end
+
+      attr_reader :job
+
+      def call_error
+        case job.exitstatus
+        when 1 # error in query sequence or options; see [1]
+          efile = File.open(job.efile)
```
yeban: Simplify this code:
- Use File.readlines to read all lines at once.
- Use Array#grep to select the lines matching ERROR_LINE.
- Show matching lines if something matched or all lines in efile.

lib/sequenceserver/blast/formatter.rb
```diff
@@ -39,10 +43,16 @@ def run
...
-          " -out '#{file}' 2> /dev/null"
+          " -out '#{file}' 2> #{efile.path}"
         logger.debug("Executing: #{command}")
         Dir.chdir(File.exist?(DOTDIR) && DOTDIR || Dir.pwd) do
           system(command)
+          status = $CHILD_STATUS.exitstatus
+          err = "Error in report generation. Maybe it's because sequence_ids are
+          not unique across the database. See -
+          https://groups.google.com/d/msg/sequenceserver/QbE11EMQoEo/nAcCOmlsSjwJ
+          Also the STDERR file: #{efile.read}"
```
vivekiitkgp: Perhaps use a HERE doc to make it more readable.

lib/sequenceserver/blast/job.rb
```diff
@@ -47,36 +48,14 @@ def run
...
-          efile.open
-          error = efile.read
-          efile.close
-          fail RuntimeError.new(status, error)
-        end
-
-        success!
+        fail RuntimeError(status, "BLAST process resulted in empty rfile.") if File.zero?(rfile)
+        done!(status)
+        success! if status == 0 && File.exist?(rfile) && !File.zero?(rfile)
```
yeban: Let's do away with `Job#success!`. Instead,

1. Define `Job#success?` as `!!exitstatus`.
2. Override it in `BLAST::Job` with the above conditional (status == 0, file exists, not empty, etc.).

lib/sequenceserver/job.rb
```diff
@@ -88,6 +88,10 @@ def success?
       !!@success
     end
 
+    def exitstatus
```
yeban: Define it as attr_reader alongside id and completed_at.

lib/sequenceserver/job.rb
```diff
@@ -109,10 +113,11 @@ def store(key, value)
       filename
     end
 
-    # Marks the job as done.
-    def done!
+    # Marks the job as done and save its exitstatus
+    def done!(status)
       @completed_at = Time.now
       @done = true
```
yeban: Let's do away with setting `@done = true`. Instead update `Job#done?` to `!!exitstatus`.

yeban:
Are you able to reproduce #194 and verify that the changes to Formatter will show a suitable error message to user?

yeban:
Can you verify if this shows an error message to the user if the rfile couldn't be written to disk for some reason? (e.g., incorrect permission on ~/.sequenceserver or low disk space). If so, is the error message suitable?

yeban:
Will our check for empty error file work in case of low disk space? I mean, we may end up with a truncated error file instead of empty error file - what then?

yeban:
The part where job is looked up by jid in routes.rb. It's possible that a job may not exist, right? e.g., because the job dir was deleted by job_cleaner.rb (or explicitly by user). Can you update the PR to show a suitable error message if a job can't be found? This is relevant because user's may bookmark link to the results / share with colleagues and come back to it after couple days and the job may have been deleted by then.

lib/sequenceserver/blast/constants.rb
```diff
@@ -2,9 +2,10 @@
 module SequenceServer
   # Constant for denoting the path ~/.sequenceserver
   DOTDIR = File.expand_path('~/.sequenceserver')
+
```
yeban: What changed here?

lib/sequenceserver/blast/job.rb
```diff
@@ -47,36 +48,12 @@ def run
...
-          efile.open
-          error = efile.read
-          efile.close
-          fail RuntimeError.new(status, error)
-        end
-
-        success!
+        done!(status)
       end
+
```
yeban: What changed here?

lib/sequenceserver/blast/job.rb
```diff
@@ -3,6 +3,7 @@
 
 require_relative 'exceptions'
 require_relative 'constants'
+require_relative 'error'
```
yeban: I think you want to require error in blast/report.rb, not here.

lib/sequenceserver/report.rb
```diff
@@ -23,6 +23,7 @@ def generate(job)
 
     def initialize(job)
       @job = job
+      BLAST::Error.new(job) unless job.success?
```
yeban: This should be in blast/report.rb.

lib/sequenceserver/routes.rb
```diff
@@ -132,12 +132,19 @@ class Routes < Sinatra::Base
       erb :'400', :layout => nil, :locals => { :error => error }
     end
 
+    # This will catch error occurred during executing blast or blast_formatter
```
yeban: How about creating a helper method, that takes the HTTP status to return and the template to render and update all error handlers to use this method?

lib/sequenceserver/job.rb
```diff
@@ -26,7 +25,9 @@ def create(params)
       #
       # TODO: What if the given job id does not exist?
       def fetch(id)
-        YAML.load_file(File.join(DOTDIR, id, 'job.yaml'))
+        job_file = File.join(DOTDIR, id, 'job.yaml')
+        fail BLAST::ArgumentError, "KeyError: Requested job was not found" unless File.exist?(job_file)
```
yeban: `BLAST::ArgumentError` is an incorrect error class for this scenario. You could raise `KeyError` here or define a new error class - I think former is okay. The corresponding handler in routes.rb should return 404.

public/js/report.js
```diff
@@ -1170,6 +1170,12 @@ var Report = React.createClass({
             case 200:
                 this.setState(jqXHR.responseJSON);
                 break;
+            case 404:
```
yeban: Use fallthrough for 404, 400 and 500.

lib/sequenceserver/blast/report.rb
```diff
@@ -23,6 +24,7 @@ module BLAST
     class Report < Report
       def initialize(job)
         super do
+          BLAST::Error.new(job) unless job.success?
```
yeban: Remove `BLAST::` from here.

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,33 @@
...
+module SequenceServer
+
+  #BLAST module
+  module BLAST
+    class Error
+      def initialize(job)
+        @job = job
+        call_error
+      end
+
```
yeban: Alias `new` to `check`.

lib/sequenceserver/blast/report.rb
yeban: In reference to [comment](https://github.com/wurmlab/sequenceserver/pull/275/files#r66506418) call here as `Error.check(job)`.

lib/sequenceserver/routes.rb
```diff
@@ -50,6 +50,14 @@ class Routes < Sinatra::Base
       }
     end
 
+    helpers do
+      def return_error(stat, erb)
```
yeban: Rename stat to code and erb to template.

lib/sequenceserver/blast/formatter.rb
```diff
@@ -24,7 +24,11 @@ def initialize(archive_file, type)
       attr_reader :format, :mime, :specifiers
 
       def file
-        @file ||= Tempfile.new filename
+        @file = File.join(File.dirname(archive_file), filename)
```
yeban: `||=`

lib/sequenceserver/blast/job.rb
```diff
@@ -47,35 +47,10 @@ def run
         logger.debug("Executing: #{command}")
 
         system("#{command} > #{rfile} 2> #{efile}")
-        done!
 
         # Capture error.
```
yeban: Remove this comment.

lib/sequenceserver/blast/exceptions.rb
```diff
@@ -1,5 +1,9 @@
 module SequenceServer
   module BLAST
+    # To signal error when Job not found
```
yeban: Use KeyError directly, without subclassing it under BLAST namespace.

lib/sequenceserver/blast/exceptions.rb
```diff
@@ -1,4 +1,7 @@
 module SequenceServer
+  # To signal error when Job not found
```
yeban: ???

lib/sequenceserver/blast/job.rb
```diff
@@ -47,35 +47,9 @@ def run
         logger.debug("Executing: #{command}")
 
         system("#{command} > #{rfile} 2> #{efile}")
-        done!
```
yeban: Remove blank line.

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'constants'
+require_relative 'exceptions'
+
+module SequenceServer
+
+  #BLAST module
```
yeban: This line is bogus.

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'constants'
+require_relative 'exceptions'
+
+module SequenceServer
+
```
yeban: No blank line here.

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,36 @@
...
+  #BLAST module
+  module BLAST
+    class Error
```
yeban: Add a one-line document.

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,36 @@
...
+      class << self
+        alias_method :check, :new
+      end
+      def initialize(job)
```
yeban: Add a blank line before initialize.

lib/sequenceserver/blast/error.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'constants'
+require_relative 'exceptions'
+
+module SequenceServer
+  module BLAST
+    class Error
+      # Catches errors encountered during the search
```
yeban: Comments usually precede the class, module, method, expression they are meant for.

lib/sequenceserver/job.rb
```diff
@@ -109,16 +110,10 @@ def store(key, value)
       filename
     end
 
-    # Marks the job as done.
-    def done!
+    # Marks the job as done and save its exitstatus
```
yeban: fullstop

lib/sequenceserver/blast/job.rb
```diff
@@ -47,35 +44,8 @@ def run
...
-          efile.open
-          error = efile.read
-          efile.close
-          fail RuntimeError.new(status, error)
-        end
-
-        success!
+        done!(status)
       end
       # rubocop:enable Metrics/AbcSize, Metrics/CyclomaticComplexity
```
yeban: Did you run rubocop / rake codeclimate to see if this comment is valid now?

yeban:
Thanks for all the great work @aniarya82 !

PR #278, opened 2016-06-10
result URLs not expected to contain hit identifiers

uludag:
If users have previously selected any hits through links in result page summary tables, their browsers history may contain URLs with  fragment identifiers for specific hits. If such links are later selected to reload the result pages we need to make sure that URLs for retrieving the results in json format are still correct.


yeban:
Thanks!

Issue #273, opened 2016-05-18
General improvements of all visualisations.


yeban:
**Alignment overview**
- [x] Tooltips are broken - they show query_m_hit_n instead of hit id.
- [x] Accomodate more hits at once in graphical overview.
- [x] #224.

**Length distribution**
- [x] Title should be left-aligned and all caps (h5/h6).
- [x] Redraw when browser window is resized.
- [x] Allow SVG and PNG download.

**Kablammo**
- [x] Kablammo polygons could do with HSP numbers, so that it's clear which polygon corresponds to which pairwise-alignment.
- [x] Redraw when browser window is resized.
- [x] No need to highlight the alignment when a polygon is selected. Numbering is enough for now.
- [x] Spacing between the polygons and axes is inconsistent.

**Circos**
- [x] Clicking anywhere should un-select the selected karyotypes.

**All**
- [x] Highlight hovered alignments in alignment overview like in length distribution (orange red colour).
- [x] Tooltips on the graphs should appear at once / without delay.
- [x] Use CSS to hide/show download buttons.

PR #281, opened 2016-06-17
Added support for Length Visualisation.

aniarya82:
Using d3's histogram added a plot to display frequency of lengths of sequences in
the hits. Plot also displays relationship between Length of sequences and query
sequence. Simailar to the Gradient Overview Length Visualisation also supports
color gradient based on the E-Values.
Here is an example:
length-vis3

### DIFF
--- a/public/js/length_visualisation.js
+++ b/public/js/length_visualisation.js
@@ -1,306 +1,247 @@
-"use strict"
+import _ from 'underscore';
+import * as Helpers from './visualisation_helpers';
 
-function LengthVisualisation(query, svg_container) {
+export default class LengthVisualisation {
+  constructor(query, $svg_container, algorithm) {
+    this._svg_container_d3 =  d3.select($svg_container[0]);
+    this.query = query;
+    this._seq_type = Helpers.get_seq_type(algorithm);
 
-  this._svg_container_d3 =  d3.select(svg_container[0]);
-  this._svg = {};
-  this._query = query;
-  this._hits = query.hits;
-  this._visualisation_helper = new VisualisationHelpers();
+    this._margin = {top: 30, right: 60, bottom: 70, left: 25};
+    this._width = $svg_container.width() - this._margin.left - this._margin.right;
+    this._height = 400 - this._margin.top - this._margin.bottom;
 
-  this._margin = {top: 30, right: 35, bottom: 50, left: 25};
-  this._width = 838 - this._margin.left - this._margin.right;
-  this._height = 460 - this._margin.top - this._margin.bottom;
-
-  this._data = this._fill_data();
-  this._x_scale = d3.scale.linear()
-      .domain([
-        d3.min([this._round_to_floor(this._query.length),this._round_to_floor(d3.min(this._data))]),
-        d3.max(this._data)
-      ])
-      .range([0, this._width]);
-  this._bins = this._create_bins();
-  this._data2 = this._fill_data2();
-  this._y_scale = d3.scale.linear()
-      .domain([0, d3.max(this._bins, function(d) { return d.length })])
-      .range([this._height, 0]);
-  // console.log("Bin details (length) "+this._bins.length);
-  // console.log('new data2 trails (length) '+this._data2.length);
-  // this._data2.map(function (d) {
-  //   console.log('data2-1 length '+d.length);
-  //   console.log('its content '+d.toString());
-  // })
-  this._axis = this._create_axis();
-  this._draw_visualisation();
-}
-
-LengthVisualisation.prototype._fill_data = function () {
-  var data = [];
-  this._hits.map(function (d) {
-    data.push(d.length)
-  })
-  // for (var i in this._hits) {
-  //   data.push(hits[i].length);
-  // }
-  return data;
-}
-
-LengthVisualisation.prototype._fill_data2 = function () {
-  var data2 = [];
-  var eScale = d3.scale.linear()
-      .domain(this._hits.map(function(d) { return d.length; }))
-      .range(this._hits.map(function (d) { return d.evalue; }));
-  this._bins.map(function (d) {
-    var inner_data = [];
-    d.map(function (e) {
-      var item = {length: e, evalue: eScale(e)};
-      inner_data.push(item);
-    })
-    data2.push(inner_data);
-  })
-  return data2;
-}
-
-LengthVisualisation.prototype._get_evalue = function (bin) {
-  var evals = [];
-  for (var i in bin) {
-    for (var j in this._hits) {
-      if (this._hits[j].length == bin[i]) {
-        evals.push(this._hits[j].evalue);
-      }
-    }
-  }
-  return evals.sort();
-}
-
-LengthVisualisation.prototype._create_formatter = function () {
-  var ticks = this._x_scale.ticks();
-  var prefix = d3.formatPrefix(ticks[ticks.length - 1]);
-  return function (d) {
-    return (prefix.scale(d) + ' ' + prefix.symbol);
-  };
-}
-
-LengthVisualisation.prototype._round_to_floor = function (d) {
-  return Math.floor(d/100)*100;
-}
-
-LengthVisualisation.prototype._create_bins = function () {
-  var bins = d3.layout.histogram()
-      .range(this._x_scale.domain())
-      .bins(this._x_scale.ticks(50))
-      (this._data);
-  return bins;
-}
-
-LengthVisualisation.prototype._get_colors_for_evalue = function (evalue) {
-  var colors = d3.scale.log()
-      .domain([
-        d3.min([1e-5, d3.min(this._hits.map(function (d) {
-          if (parseFloat(d.evalue) === 0) return undefined;
-          return d.evalue;
-        }))]),
-        d3.max(this._hits.map(function (d) {
-          return d.evalue;
-        }))
-      ])
-      .range([40,150]);
-  var rgb = colors(evalue);
-  return d3.rgb(rgb,rgb,rgb);
-}
-
-LengthVisualisation.prototype._set_gradient_data = function (bin) {
-  var set = [];
-  var eValue = this._get_evalue(bin);
-  var offsetScale = d3.scale.linear()
-      .domain([0,eValue.length])
-      .range([0,100]);
-  var color_init = this._visualisation_helper._get_colors_for_evalue(eValue[0],this._hits);
-  set.push({offset: '0%', color: color_init});
-  for (var i in eValue) {
-    var j = parseInt(i)+1;
-    var offset = offsetScale(j)+'%';
-    var rgb = this._visualisation_helper._get_colors_for_evalue(eValue[i],this._hits);
-    var item = { offset: offset , color: rgb };
-    set.push(item);
-  }
-  return set;
-}
-
-LengthVisualisation.prototype._draw_visualisation = function () {
-  this._svg.d3 = this._svg_container_d3.insert('svg', ':first-child')
-      .attr('width', this._width + this._margin.right + this._margin.left)
-      .attr('height', this._height + this._margin.top + this._margin.bottom)
-      .append('g')
-      .attr('transform','translate('+this._margin.left+','+this._margin.top+')');
-
-  this._draw_rectangles();
-  this._draw_query_line();
-  this._draw_scales();
-  this._create_labels();
-  this._draw_legend();
-}
-
-LengthVisualisation.prototype._draw_rectangles = function () {
-  var self = this;
-  var getEvalue = function(bin) {
-    for (var i in self._hits) {
-      if (self._hits[i].length == d3.min(bin)) {
-        return self._hits[i].evalue;
-      }
-    }
-  }
-  var bar = this._svg.d3.selectAll('.bar')
-      .data(this._bins)
-      .enter().append('g')
-      .attr('class', 'bar')
-      .attr('data-toggle','tooltip')
-      .attr('title', function (d) {
-        return '['+d.toString()+']'+'('+self._y_scale(d.length)+')';
-      })
-      .attr('transform', function(d) {
-        return 'translate('+(self._x_scale(d.x)+self._margin.left)+','+self._y_scale(d.length)+')';
-      });
-
-  // bar.selectAll('rect')
-  //     .data(function (d) { return d; })
-  //     .enter().append('rect')
-  //     .attr('x', 1)
-  //     .attr('y', function(i) { return self._y_scale(d.length); })
-  //     .attr('width', self._x_scale(this._bins[1].x) - self._x_scale(this._bins[0].x) - 1)
-  //     .attr('height', function (i) { return self._height - self._y_scale(d.length); })
-  //     .attr('fill', self._get_colors_for_evalue(getEvalue(d)));
-
-  bar.append('rect')
-      .attr('x', 1)
-      .attr('width', self._x_scale(this._bins[1].x) - self._x_scale(this._bins[0].x) - 1)
-      .attr('height', function(d) { return self._height - self._y_scale(d.length); })
-      .style('fill',function (d) {
-        // return self._get_colors_for_evalue(getEvalue(d));
-        return 'url(#gradient'+d[0]+')';
-      });
-
-  this._bins.map(function (bin) {
-    if (bin.length != 0) {
-      var grad = self._svg.d3.append('linearGradient')
-          .attr('id','gradient'+bin[0])
-          .attr('x1','0%').attr('y1','0%')
-          .attr('x2','0%').attr('y2','100%')
-          .selectAll('stop')
-          .data(self._set_gradient_data(bin))
-          .enter().append('stop')
-          .attr('offset', function (e) { return e.offset; })
-          .attr('stop-color', function (e) { return e.color; });
-    }
-  });
-}
-
-LengthVisualisation.prototype._draw_query_line = function () {
-  var self = this;
-  var query_line = this._svg.d3.append('g')
-      .attr('class','query_line')
-      .attr('transform','translate('+(this._margin.left+this._x_scale(this._query.length))+',0)');
-
-  query_line.append('rect')
-      .attr('x',1)
-      .attr('width',4)
-      .attr('height',self._height)
-      .style('fill','rgb(102,102,102)');
-
-  query_line.append('text')
-      .attr('dy', '0.75em')
-      .attr('y', -5)
-      .attr('x', 5)
-      .attr('text-anchor','start')
-      .text('Query')
-      .style('fill','#000')
-      .attr('transform','rotate(-45)');
-}
-
-LengthVisualisation.prototype._create_axis = function () {
-  var formatter = this._create_formatter()
-  var format = d3.format('.0g');
-  var x_axis = d3.svg.axis()
-      .scale(this._x_scale)
-      .orient('bottom')
-      .ticks(50)
-      .tickFormat(formatter);
-
-  var y_axis = d3.svg.axis()
-      .scale(this._y_scale)
-      .orient('left')
-      // .ticks(20)
-      // .tickFormat(format);
-
-  return {x: x_axis, y: y_axis};
-}
-
-LengthVisualisation.prototype._draw_scales = function () {
-  var self = this;
-  var xContainer = this._svg.d3.append('g')
-      .attr('class', 'axis axis--x')
-      .attr('transform', 'translate('+self._margin.left+','+self._height+')')
-      .call(this._axis.x);
-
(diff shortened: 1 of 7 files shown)


yannickwurm:
Thanks v much Hiten - this is a step in the right direction - clearly still many shortcomings - please email me when you have a more mature concept

cheers,
Yannick

aniarya82:
An updated version
- Fixed the ticks formatting in x-axis.
- Moved legend to bottom of the visualisation.
- Reduced the bins sizes to make graph more precise.
- Changed the format of length-visualasation.js file to make it Object oriented.
screen shot 2016-06-23 at 2 10 57 pm

public/js/report.js
```diff
@@ -742,6 +743,25 @@ var Hit = React.createClass({
 });
 
 /**
+ * Renders Length Visualisation of all hits per query
```
yeban: I guess length visualisation could be of many type - length distribution would a more appropriate name.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,306 @@
+"use strict"
+
+function LengthVisualisation(query, svg_container) {
```
yeban: Why does every function and property begin with `_`?

yeban: Why not use ES6 classes instead?

public/js/report.js
```diff
@@ -742,6 +743,25 @@ var Hit = React.createClass({
...
+ */
+var LengthVis = React.createClass({
+  mixins: [Utils],
```
yeban: It's not used here, is it?

public/js/visualisation_helpers.js
```diff
@@ -0,0 +1,65 @@
+function VisualisationHelpers() {
```
yeban: Use [ES6 modules](http://exploringjs.com/es6/ch_modules.html) instead.

public/js/visualisation_helpers.js
```diff
@@ -0,0 +1,65 @@
...
+  var uniq = [];
+  for(var i = 0; i < arr.length; i++) {
+    while(i < (arr.length - 1) && arr[i] === arr[i + 1]) { i++; }
+    uniq.push(arr[i]);
+  }
+  return uniq;
+}
+
+/**
+ * Defines how ticks will be formatted.
```
yeban: Please could you add an explanation of what this does with 2-3 examples?

public/js/visualisation_helpers.js
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * Modified by Priyam based on https://github.com/mbostock/d3/issues/1722.
```
yeban: Now that it's out of kablammo folder, better to say - "borrowed from kablammo - modified by ..."

public/js/visualisation_helpers.js
```diff
@@ -0,0 +1,65 @@
...
+ */
+VisualisationHelpers.prototype._create_formatter  = function (scale, seq_type) {
```
yeban: Call it tick_formatter?

public/js/visualisation_helpers.js
```diff
@@ -0,0 +1,65 @@
...
+        }))
+      ])
+      .range([40,150]);
+  var rgb = colors(evalue);
+  return d3.rgb(rgb,rgb,rgb);
+}
+
+// Returns copy of `arr` containing only unique values. Assumes duplicate
+// values always occur consecutively (which they will if `arr` is sorted).
+VisualisationHelpers.prototype._uniq = function (arr) {
```
yeban: Please use [`_.uniq`](http://underscorejs.org/#uniq) instead of this helper method.

vivekiitkgp:
@aniarya82 Let me know if I am missing something. Since the visualization above is a histogram, the function is essentially binning the data and showing the distribution. Which means that by controlling the number of bins to use, you can make the distribution smooth or rough. Did you experiment with this? What's the currently used strategy?

vivekiitkgp:
@aniarya82 Also, I think it'd be handy to have a hover tooltip for histogram bars that shows the number of HSPs, Identity and Query coverage values.

yannickwurm:
- [ ] start X axis with 0, end it with 1.05*max(hits, query)  (to have extra space on either side).
 - [ ] every hit's must have a different color (right now a bit has the color of the minimum evalue  - this can be misleading
 - [ ] use less vertical space
 - [ ] query should be more visible
 - [ ] too many lablels. Perhaps can have fewer ticks:use (perhaps) between 4 labels and max 12 ticks in total? (like kablammo). And/or not every tick needs a label.
 - [ ] keep 1.0 rather than 1 (if also showing e.g. 0.8 or 0.5)
 - [ ] add bp and aa when appropriate. 
 - [ ] Consider whether we should shift bins by 50% of tick width - like [here](https://learnandteachstatistics.files.wordpress.com/2012/11/histogram_0.png)
 - [ ] no need for a separate legend. 
 - [ ] beginning and end of x axis should be vertically aligned with first and last ticks of overview graph. 
 (no need for funky legend colors because we have a single continuous variable with no biologically meaningful transitionpoints).

aniarya82:
Changes in length distribution chart:
- Reduced the vertical space taken up by chart
- changes x-axis domain ([0,1.05*max(hits,query)])
- altered the color of the query line to make it more legible
- reduced the number of ticks labelled
- added 'aa' and 'bp' when appropriate
- removed the legend
length-vis

yannickwurm:
Glad to see this is possilble :)
Can you tick the checkboxes in the summary of our discussion above? 
Also:
- [ ] flip the colors so the strongest is at the bottom
- [ ] no "k" abbreviation for amino acids

Cheers!
y

yeban:
This looks great @aniarya82! We shouldn't include 'aa' / 'bp' for 0. Could you also try to show tooltips indicating hit name, length and evalue? Can we have the same tooltip generation code reused in both graphical overview and length visualisation?

@yannickwurm Proteins can also get mighty big (e.g., actin and all). The tick formatter used here is adaptive: it's when the numbers get big that they are reformatted with SI prefix - 'k', 'm', 'g', for 10^3, 10^6, and 10^9. We inherited the technique from Kablammo and use it in graphical overview too (one of the other things in master). Instead of removing 'k' for 'aa', how about rendering '2.5k aa' instead of '2.5 kaa'. Or, we don't do 'k' SI prefix till 50,000 for both aa and bp (knowing that the largest protein we know of is 33,000 aa long).

aniarya82:
Few updates:
- Flipped colours, now strongest hit is at bottom
- Removed 'k' abbreviation for amino acids
- Removed 'aa' / 'bp' for 0
- Added tooltips indication hit name, evalue and length.
length-vis 2

yeban:
I just tried this locally. The y-axis looks like in one of the first pics you posted:

screen shot 2016-06-30 at 4 35 28 pm

yeban:
I think the tooltips should be displayed without a delay.

aniarya82:
Minor updates:
- Y-axis formatting fixed.
- reduced tooltips delay to 0.
screen shot 2016-06-30 at 8 07 17 pm

yeban:
- [ ] It is not very meaningful to have ticks on Y axis apart from 1, 2, 3, N.
- [ ] I think we should show only one 0 for both axes combined.
- [ ] Can you align it with graphical overview?

public/js/visualisation_helpers.js
```diff
@@ -0,0 +1,57 @@
+import _ from 'underscore';
+
+export default class VisualisationHelpers {
```
yeban: We don't want a class here. Just a collection of methods in a file. Import in graph.js or length_visualisation.js as `import * as Helpers from './visualisation_helpers.js'` and use as `Helpers.tick_formatter`.

public/js/report.js
```diff
@@ -742,6 +745,24 @@ var Hit = React.createClass({
...
+var LengthVis = React.createClass({
+  plot_svg: function () {
+    return $(React.findDOMNode(this.refs.plot_test));
+  },
+  componentDidMount: function () {
+    this.plot_test = new LengthVisualisation(this.props.query, this.plot_svg(), this.props.algorithm);
+  },
+  render: function () {
+    return(
+      <div ref="plot_test">
```
yeban: I would prefer the name "svgContainer". Also, give it a class: length-distribution.

public/js/report.js
```diff
@@ -742,6 +745,24 @@ var Hit = React.createClass({
...
 
 /**
+ * Renders Length Distribution of all hits per query
+ */
+var LengthVis = React.createClass({
+  plot_svg: function () {
+    return $(React.findDOMNode(this.refs.plot_test));
+  },
+  componentDidMount: function () {
+    this.plot_test = new LengthVisualisation(this.props.query, this.plot_svg(), this.props.algorithm);
```
yeban: Store the object in `this.graph`.

public/js/report.js
```diff
@@ -742,6 +745,24 @@ var Hit = React.createClass({
 });
 
 /**
+ * Renders Length Distribution of all hits per query
+ */
+var LengthVis = React.createClass({
+  plot_svg: function () {
```
yeban: Rename it to `svgContainer` in line with GraphicalOverview and Kablammo components.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+  constructor(query, svg_container, algorithm) {
+    this._svg_container_d3 =  d3.select(svg_container[0]);
+    this._svg = {};
+    this._query = query;
+    this._seq_type = this.get_seq_type(algorithm);
+    this._hits = query.hits;
+    this.visualisation_helpers = new VisualisationHelpers();
+
+    this._margin = {top: 30, right: 60, bottom: 70, left: 25};
+    this._width = 838 - this._margin.left - this._margin.right;
```
yeban: Shouldn't it be svg_container.width() instead of 838 here?

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
+import _ from 'underscore';
+import VisualisationHelpers from './visualisation_helpers';
+
+export default class LengthVisualisation {
+  constructor(query, svg_container, algorithm) {
+    this._svg_container_d3 =  d3.select(svg_container[0]);
+    this._svg = {};
+    this._query = query;
```
yeban: `_` is used to indicate that this property is private. We don't need to do that for query.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+    this._seq_type = this.get_seq_type(algorithm);
+    this._hits = query.hits;
```
yeban: Redundant - you can always do `this.query.hits`.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+    this._hits.map(function (d) {
+      data.push(d.length)
+    })
+    // for (var i in this._hits) {
+    //   data.push(hits[i].length);
+    // }
+    return data;
+  }
+
+  get_seq_type(algorithm) {
```
yeban: Shouldn't this be moved to helpers?

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+    this._svg.d3 = this._svg_container_d3.insert('svg', ':first-child')
+        .attr('width', this._width + this._margin.right + this._margin.left)
+        .attr('height', this._height + this._margin.top + this._margin.bottom)
+        .append('g')
+        .attr('transform','translate('+this._margin.left+','+this._margin.top+')');
+    this.setupTooltip();
+    this.draw_visualisation();
+  }
+
+  fill_data() {
```
yeban: This is not a meaningful name. Should be `hit_lengths`.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+    var data = [];
```
yeban: `map` returns an array. You are discarding `map`'s return value and using it like `each`.

`return _.map(this.query, _.iteratee('length'))`

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+      })
+      var item = {data: inner_data, x: d.x, dx: d.dx, length: d.length};
+      data2.push(item);
+    })
+    return data2;
+  }
+
+  draw_rectangles() {
+    var self = this;
+    var getEvalue = function(bin) {
```
yeban: Unused function it seems?

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+    $('[data-toggle="tooltip"]').tooltip({
+      'placement': 'top',
+      'container': 'body',
+      'html': 'true',
+      'delay': 0,
+      'white-space': 'nowrap'
+    });
+  }
+
+  get_evalue(bin) {
```
yeban: Unused function?

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+        .attr('y', -5)
+        .attr('x', 5)
+        .attr('text-anchor','start')
+        .text('Query')
+        .style('fill','#000')
+        .attr('transform','rotate(-45)');
+  }
+
+  create_axis() {
+    var formatter = this.tick_formatter(this._seq_type.subject_seq_type);
```
yeban: Why are we not using `tick_formatter` from visualisation helpers?

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+    var power = match[3];
+    var prettyEvalue = parseFloat(base).toFixed(2);
+    var returnString = '<br>E value: ' + prettyEvalue;
+    if (power !== undefined) {
+        returnString +=  ' &times; 10<sup>' + power + '</sup>';
+    }
+    return returnString;
+  }
+
+  fill_data2() {
```
yeban: This function is very unclear - including names like data2, reusing the same variable name in different contexts, a method called set_evalue that matches based on length.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+        .attr('dy', '0.75em')
+        .attr('y', -5)
+        .attr('x', 5)
+        .attr('text-anchor','start')
+        .text('Query')
+        .style('fill','#000')
+        .attr('transform','rotate(-45)');
+  }
+
+  create_axis() {
```
yeban: Creates both x and y axis. Should be more appropriately named - create_axes.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+
+    this._svg.d3.append('text')
+        .attr('class', 'yaxis-label')
+        .attr('x',-255)
+        .attr('y',-15)
+        .attr('transform','rotate(-90)')
+        .text('Number of Sequences');
+  }
+
+  draw_legend() {
```
yeban: Is this the exact same code from graphical overview? If so it's safe to remove it I guess since we are not drawing a legend here.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
+import _ from 'underscore';
+import VisualisationHelpers from './visualisation_helpers';
+
+export default class LengthVisualisation {
+  constructor(query, svg_container, algorithm) {
```
yeban: Let's call it `$svg_container` to be clear that it's jquery wrapped DOM node.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+    this._svg_container_d3 =  d3.select(svg_container[0]);
+    this._svg = {};
```
yeban: `this._svg` is an object with only one property (`d3` - L36). And `this._svg.d3` is same as `this._svg_container_d3`. Right? Can we have just one `this.svg` instead?

public/js/length_visualisation.js
```diff
@@ -0,0 +1,339 @@
...
+    this._data = this.fill_data();
+    this._x_scale = d3.scale.linear()
+        .domain([
+          0,
+          (d3.max([this._query.length,d3.max(this._data)]) * 1.05)
+        ])
+        .range([0, this._width]);
+    this._bins = this.create_bins();
+    this._data2 = this.fill_data2();
+    this._y_scale = d3.scale.linear()
```
yeban: `this._y_scale` is used only once in `create_axis`. Why not define the variable in `create_axis` method?

yannickwurm:
- [ ] labels are not fully centered under tickmarks
 - [ ] unlabelled ticks should be lighter or thinner or shorter than those with labels 
 - [ ] y axis label should be vertically centered with regards to Y axis. Actually no: remove axis labels and instead use an informative overall title for the graph. This will allow the width of the x axis to be identical to the graph above. 
 - [ ] a rounding error (or similar) seems to introduce white lines between hit boxes in some cases. And to make some ticks thicker than others. Perhaps activating antialiasing would resolve this?
 - [ ] in old overview visualisation, try putting 0px or 1px between hits so as to vertically compress results.
 - [ ] vertically take up only 30% of current space
 - [ ] check what standard practice is in other beuatiful histogram and bar charts regarding where the axes end. If apporpriate, do what they do (e.g. extend the y and x axis to the next labelled tick).

yannickwurm:
-[ ] if there is a good code checker (for syntax, accuracy & svg integrity) please use it.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,247 @@
...
+    if (power !== undefined) {
+        returnString +=  ' &times; 10<sup>' + power + '</sup>';
+    }
+    return returnString;
+  }
+
+  update_data() {
+    var self = this;
+    var data2 = [];
+    var set_index = function(length) {
```
yeban: There's probably a function in underscorejs to do this.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,247 @@
...
+import * as Helpers from './visualisation_helpers';
+
+export default class LengthVisualisation {
+  constructor(query, $svg_container, algorithm) {
+    this._svg_container_d3 =  d3.select($svg_container[0]);
+    this.query = query;
+    this._seq_type = Helpers.get_seq_type(algorithm);
+
+    this._margin = {top: 30, right: 60, bottom: 70, left: 25};
+    this._width = $svg_container.width() - this._margin.left - this._margin.right;
```
yeban: `width` and `height` are used only inside the constructor. They thus don't need to be a property.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,247 @@
...
+        y_axis.tickValues(d3.range(0, d3.max(this._bins, function(d) { return d.length })+1));
+        break;
+      }
+    }
+
+    return {x: x_axis, y: y_axis};
+  }
+
+  draw_scales() {
+    var self = this;
```
yeban: Doesn't look like assigning `this` to `self` is needed here.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,247 @@
...
+        .attr('transform','rotate(-90)');
+
+    var yContainer = this.svg.append('g')
+        .attr('class','axis axis--y')
+        .attr('transform','translate('+self._margin.left+',0)')
+        .call(this._axis.y);
+  }
+
+  create_labels() {
+    var self = this;
```
yeban: `self = this` not needed here either, I think.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,247 @@
...
+
+  create_bins() {
+    var bins = d3.layout.histogram()
+        .range(this._x_scale.domain())
+        .bins(this._x_scale.ticks(50))
+        (this._data);
+    return bins;
+  }
+
+  prettify_evalue(evalue) {
```
yeban: This functionality is needed in at least 4 places that I can remember: tabular summary, alignment overview (tooltip), stats table before pairwise alignment. And now length distribution. Could you extract this into a helper method and use the same throughout. Treat `report.js`' `prettify_float` to be authoritative.

Similarly, tooltip generation should be shared between this and alignment overview.

public/js/length_visualisation.js
```diff
@@ -0,0 +1,247 @@
...
+          (d3.max([this.query.length,d3.max(this._data)]) * 1.05)
+        ])
+        .range([0, this._width]);
+    this._bins = this.create_bins();
+    this._update_data = this.update_data();
+    this._y_scale = d3.scale.linear()
+        .domain([0, d3.max(this._bins, function(d) { return d.length })])
+        .range([this._height, 0]);
+    this._axis = this.create_axes();
+    this.svg = this._svg_container_d3.insert('svg', ':first-child')
```
yeban: Doesn't the right hand side expression return `this._svg_container_d3`?

yeban:
Great work! Thanks.

Could you add anything left to take care of to #273 along with other stuff that we discussed (redraw on resize, svg & png download buttons)?

PR #282, opened 2016-07-12
Vis improvements

aniarya82:


public/css/svg.css
```diff
@@ -31,7 +47,8 @@
 }
 
 .length-distribution-title {
-  text-align: center;
+  text-align: left;
+  text-transform: capitalize;
```
yeban: I think you mean `uppercase`.

public/js/lengthdistribution.js
```diff
@@ -15,16 +15,37 @@ export default class LengthDistribution extends React.Component {
...
+    .on('mouseleave', function() {
+      $(this).find('.hit-links').hide();
+    })
   }
 
   render() {
     return(
       <div>
+
         <div className='length-distribution-title'>
```
yeban: I don't think this `div`is needed.

public/css/svg.css
```diff
@@ -22,6 +22,22 @@
   stroke-dasharray: 5,5;
 }
 
+.hit-links {
+  float: right;
+}
+
+.kablammo {
```
yeban: Put `grapher` class on the parent `div` of `ref=svgContainer` and apply this style to `.grapher` to get the effect for all visualisation instances at once.

public/js/lengthdistribution.js
```diff
@@ -15,16 +15,37 @@ export default class LengthDistribution extends React.Component {
...
+    svgContainer
+    .on('mouseover', function () {
+      $(this).find('.hit-links').show();
+    })
+    .on('mouseleave', function() {
+      $(this).find('.hit-links').hide();
+    })
   }
 
   render() {
```
yeban: I think you can take out the `render` method to a `Grapher` mixin as it will be the same for each graph (a daddy container with download buttons on top right and svg container to serve as canvas for drawing the graph.

public/js/lengthdistribution.js
```diff
@@ -88,6 +110,17 @@ export class Graph {
     });
   }
 
+  setupResponsiveness() {
```
yeban: Would be great if you can set this up as a jquery extension that adds _one_ event handler to `window.resize`. The handler looks for all `.grapher` divs, fetches the corresponding ReactComponent instance and calls `.graph.draw` on that. In a second step we can optimise redrawing of the graphs by for example, debouncing them at different intervals.

public/js/report.js
```diff
@@ -287,11 +287,24 @@ var Kablammo = (function () {
             .on('mouseenter', null)
```
yeban: Can you move Kablammo react component to kablammo.js, and appropriately require and use it in report.js?

public/js/lengthdistribution.js
```diff
@@ -15,16 +15,37 @@ export default class LengthDistribution extends React.Component {
...
     return(
       <div>
+
         <div className='length-distribution-title'>
-          <h4>Frequency of Hits length</h4>
+          <h5>Frequency of Hits length</h5>
         </div>
         <div ref="svgContainer" className='length-distribution'>
+          <div
+              className="hit-links" style={{display:"none"}}>
```
yeban: s/hit-links/graph-links

public/js/report.js
```diff
@@ -1400,7 +1432,7 @@ var Report = React.createClass({
 
     // SVG and PNG download links.
     setupKablammoImageExporter: function () {
-        new ImageExporter('#report', '.export-to-svg', '.export-to-png');
+        new ImageExporter('.subject', '.export-to-svg', '.export-to-png');
```
yeban: I think image exporter should work on grapher class.

public/css/svg.css
yeban: See if h6 doesn't already provide upercasing. If it doesn't either restyle h6 or create a class `caption` to be applied to h6 (or h5).

yeban:
Feature wise, I am not happy with the changes to the legend. There's too much space between the weakest evalue and the gradient. And the strongest evalue doesn't render as HTML.

screen shot 2016-08-01 at 1 00 15 pm

lib/sequenceserver/links.rb
```diff
@@ -116,25 +116,25 @@ def uniprot
...
-    def png_download
-      {
-        :order => 2,
-        :title => 'PNG',
-        :url   => '#',
-        :class => 'export-to-png',
-        :icon  => 'fa-download'
-      }
-    end
+    # def svg_download
```
yeban: Please remove the code, instead of commenting it out. If this code is required again, we can fetch it from git.

public/css/svg.css
```diff
@@ -22,6 +22,28 @@
   stroke-dasharray: 5,5;
 }
 
+.graph-links {
+  right: 3%;
+  position: absolute;
+  z-index: 10;
+}
+
+.svgContainer {
```
yeban: Use snake-case for CSS.

public/js/grapher.js
```diff
@@ -0,0 +1,108 @@
...
+//                 <i className="fa fa-download"/>
+//                 <span>{"  PNG  "}</span>
+//             </a>
+//         </div>
+//       </div>
+//     );
+//   }
+// }
+
+export function graph_links(svgContainer) {
```
yeban: Can't this be implemented in CSS? I'm guessing:

```
.graph-links {
  display: none;
}

.grapher:hover .graph-links {
display: block; /* or inline or whatever */
}
```

public/js/graphicaloverview.js
```diff
@@ -160,13 +160,19 @@ import * as Helpers from './visualisation_helpers';
             .attr('class',' legend-text')
             .attr('transform', 'translate(0, ' +options.legend +')')
             .attr('x', 6 * (width - 2 * options.margin) / 10 - options.margin / 2)
-            .text("Weaker hits");
+            // .text("Weaker hits");
+            .text(function() {
```
yeban: Shouldn't it be `html` here instead of `text`?

public/js/graphicaloverview.js
```diff
@@ -160,13 +160,19 @@ import * as Helpers from './visualisation_helpers';
...
+              return Helpers.prettify_evalue(hits[hits.length-1].hitEvalue);
+            })
 
         svg_legend.append('text')
             .attr('class',' legend-text')
             .attr('transform', 'translate(0, ' + options.legend + ')')
             .attr('x', 9 * (width - 2 * options.margin) / 10 + options.margin / 2)
-            .text("Stronger hits");
+            // .text("Stronger hits");
+            .text(function () {
```
yeban: Same.

PR #285, opened 2016-08-03
Circos improvements

aniarya82:


yeban:
It would be great if the colours indicated significance of the alignment. Increasing -> red, orange, blue.

yeban:
Much better. Thanks!

PR #286, opened 2016-08-09
Circos improvements

aniarya82:


aniarya82:
Snapshot of click feature for karyotypes.
screen shot 2016-08-09 at 5 57 00 am

PR #290, opened 2016-08-16
Circos labels and ticks fonts set to 11px

aniarya82:
Shift of last tick removed
Label cut-off improved


yeban:
Ticks are now left-shifted. Labels still don't use all available space.

screen shot 2016-08-16 at 1 44 42 pm

PR #306, opened 2016-09-08
[alignment] arrow added

aniarya82:


yeban:
I had something like this in mind:

feature

Issue #309, opened 2016-09-30
won't find blast databases witha filename ending in a number


guyleonard:
The database scan seems to ignore / not find dbs ending in a number: e.g "rhizaria80.nal" is ignored but "rhizaria80s.nal" will be found.

yeban:
It is to ignore volumes of multi-part databases (like those downloadable from NCBI). But it does look like our regex is buggy: it's not looking for a `.` before the ending numbers.

yeban:
Ok. While the regex should look for `.` before the ending numbers, that is not the reason for your issue:

```
$ sequenceserver -i
irb#1(SequenceServer):002:0> Database.multipart_database_name? "/tmp/rhizaria80.nal"
=> false
irb#1(SequenceServer):001:0> Database.multipart_database_name? "/tmp/rhizaria80s.nal"
=> false
```

Volumes of multi-part database end in digits, this ends in `nal`.

Do you get an error output that might be of help?

guyleonard:
Indeed that makes sense, no unfortunately when I looked last week I did not see any error message whilst running in debug mode!?

guyleonard:
Just checked, looks like it is fixed! Thanks. Sorry for slow reply! :)

PR #311, opened 2016-10-04
More explicit multipart DB name regex

nathanhaigh:
This is to fix issue #309 regarding DB names with numerical suffixes being incorrectly identified as being multipart. This PR attempts to fix the issue by having a more restrictive regex which requires the multipart numerical suffix to be preceeded by a dot.

I think I also added a test for this case - hope this works!


yeban:
Thanks. +1 for test case.

PR #317, opened 2016-11-16
set query sequence from POST data

rjchallis:
We've been finding it very useful to be able send query sequences as POST data to the sequenceserver front end so users can choose databases, set parameters, blast and then link back to lepbase.  I see you have an API now so don't know if this approach is compatible with your plans but it would be a really useful feature to have in the main branch.

### DIFF
--- a/lib/sequenceserver/routes.rb
+++ b/lib/sequenceserver/routes.rb
@@ -58,6 +58,7 @@ module SequenceServer
 
     # Returns base HTML. Rest happens client-side: rendering the search form.
     get '/' do
+      @input_sequence = ''
       erb :layout
     end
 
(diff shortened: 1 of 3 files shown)


yeban:
Thanks. I think it would be useful to other lebpase like users as well. I didn't have any plans for this. But if the sequences are available on the server, which should be the case most of the time, perhaps `/searchdata.json` would be better for this. Something like:

- Load lepbase.com/blast?query=query_id_1,query_id_2
- Query params get passed on to `/searchdata.json` which additionally returns query sequences and is loaded into the text area.

rjchallis:
I felt that this approach was useful as it allows for things like sending a single exon sequence to sequenceserver without needing to have a file of individually named exon sequences. But we're not actually using it like that at the moment so using ``/searchdata.json`` is definitely a more elegant approach - thanks for the tip. 

The other approach I could envisage wanting to adopt is blasting the flanking sequence of a feature on a scaffold, so the scaffold would be available on the blast server but I'd want to use the region between specific coordinates as the query.  Any plans to allow a subsequence to be defined in a query param or would this be best achieved with POST?

yeban:
You could encode a subsequence as `query=seqid1:start-stop,seqid2:start-stop`. Sequence id goes to `-entry` option and range to `-range` option of `blastdbcmd`.

If you implement query params for `/searchdata.json`, I will happy to receive a pull request.

rjchallis:
Thank you, this looks like a great solution - I'll get back to you once I've had a chance to implement it

PR #319, opened 2016-12-20
added more colors to color palette

aniarya82:
Added more colors to make Circos better.
screen shot 2016-12-20 at 11 05 38 pm


yeban:
Thanks, @aniarya82. I had a look. I am trying to assess how significant is the improvement here. Also, some of the CSS syntax used here is new to me, e.g., `var(--m)`. What does that do? What is the browser compatibility like for this syntax?

It would be very helpful if you can provide more screenshots, showing how the visualisation changes for the same BLAST results with this PR. I think setting a higher opacity (lesser translucency), or none at all, for the ribbons will improve contrast. On a related note, the change-in-opacity-on-hover effect is probably not required since you anyway hide other ribbons on hover.

aniarya82:
Before increasing colors in color palette.
before color palette addition
After adding more colors to color palette, opacity of ribbons = 0.55
opacity 0 55
After adding more colors to color palette, opacity of ribbons = 0.85
opacity 0 85

yeban:
Thanks.

PR #322, opened 2016-12-29
New feature 'sys'.

enjoyyourcopepods:
Defined the new method 'sys' in lib/sequenceserver.rb. Added "require
'tempfile'" to lib/sequenceserver.rb for 'sys'. Created the
CommandFailed class in lib/sequenceserver/exceptions.rb for 'sys'.
Changed lib/sequenceserver/database.rb to make use of 'sys'.

Signed-off-by: Lawrence John Maynard <enjoyyourcopepods@gmx.com>

### DIFF
--- a/lib/sequenceserver.rb
+++ b/lib/sequenceserver.rb
@@ -66,21 +66,21 @@ module SequenceServer
     # 'sys' can get from a failed shell command stdout, stderr, and exit status.
     #
     # Supply 'sys' with the shell command and optionally:
-    # directory: A directory to change to for the duration of the execution of
+    # dir: A directory to change to for the duration of the execution of
     # the shell command.
     # path: A directory to change the PATH environment variable to for the
     # duration of the execution of the shell command.
-    # stdout_file: A path to a file to store stdout.
-    # stderr_file: A path to a file to store stderr.
+    # stdout: A path to a file to store stdout.
+    # stderr: A path to a file to store stderr.
     #
     # Usage:
     #
-    # stdout, stderr = sys(command, :directory => '/path/to/directory',
+    # stdout, stderr = sys(command, :dir => '/path/to/directory',
     # :path => '/path/to/directory'
     #
     # sys(command, :directory => '/path/to/directory',
-    # :path => '/path/to/directory', :stdout_file => '/path/to/stdout_file',
-    # :stderr_file => '/path/to/stderr_file'
+    # :path => '/path/to/directory', :stdout => '/path/to/stdout_file',
+    # :stderr => '/path/to/stderr_file'
 
     def sys(command, options = {})
       # Store the initial value of the PATH environment variable.
@@ -98,7 +98,7 @@ module SequenceServer
 
       logger.debug("Executing: #{command}")
 
-      directory = options[:directory] || Dir.pwd
+      directory = options[:dir] || Dir.pwd
 
       # Change the directory, execute the shell command, redirect stdout and stderr
       # to the temporary files.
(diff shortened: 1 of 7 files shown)


lib/sequenceserver.rb
```diff
@@ -57,6 +58,83 @@ def init(config = {})
...
 
+    # 'sys' executes a shell command.
+    #
+    # 'sys' can write the stdout and/or stderr from a shell command to files, or
+    #  return these values.
+    #
+    # 'sys' can get from a failed shell command stdout, stderr, and exit status.
+    #
+    # Supply 'sys' with the shell command and optionally:
+    # A directory to change to for the duration of the execution of the shell
```
yeban: Give the name of the corresponding keys (`directory: A directory to change to ...`).

lib/sequenceserver.rb
```diff
@@ -57,6 +58,83 @@ def init(config = {})
...
+    # A directory to change the PATH environment variable to for the duration of the
+    # execution of the shell command.
+    # A path to a file to write stdout to.
+    # A path to a file to write stderr to.
+    #
+    # Usage:
+    #
+    # stdout, stderr = sys(command, :directory => '/path/to/directory',
+    # :path => '/path/to/directory', :stdout_file => '/path/to/stdout_file',
+    # :stderr_file => '/path/to/stderr_file'
```
yeban: Invoking this way doesn't return anything, right?

lib/sequenceserver.rb
```diff
@@ -57,6 +58,83 @@ def init(config = {})
...
+        system("#{command} 1>#{temp_stdout_file.path} 2>#{temp_stderr_file.path}")
+      end
+
+      unless $CHILD_STATUS.success?
+        raise CommandFailed.new(temp_stdout_file.read, temp_stderr_file.read, $CHILD_STATUS.exitstatus)
+      end
+
+      # Write stdout and/or stderr to files, if paths for them were given.
+      # If a full path was given for an output file, write the contents of
+      # the temporary file to this path. If the path given does not
```
yeban: We don't write the tempfiles to the given location, we move it (it's a feature, hence worth highlighting).

lib/sequenceserver.rb
```diff
@@ -57,6 +58,83 @@ def init(config = {})
...
+          unless File.directory?(file_dir)
+            FileUtils.mkdir_p(file_dir)
+          end
+          FileUtils.mv([temp_stdout_file, temp_stderr_file][index], filename)
+        end
+      end
+
+      # If paths to write stdout and/or stderr to were not given, return the
+      # contents of stdout and/or stderr. Otherwise, return nil.
+      return_array = []
```
yeban: I gave it some thought. Let's return if neither `:stdout_file` nor `:stderr_file` was specified.

lib/sequenceserver/database.rb
```diff
@@ -113,16 +113,19 @@ def clear
...
+          out, err = sys(cmd)
+        rescue
+          fail BLAST_DATABASE_ERROR.new(cmd, err)
+        end
+        out.each_line do |line|
+          name = line.split('	')[0]
+          next if multipart_database_name?(name)
+          self << Database.new(*line.split('	'))
         end
+        errpat = /BLAST Database error/
```
yeban: - Should check for errors and raise before parsing the output.
- Could you move running of the command and failure scenarios to a new method? Call it `blastdbmcmd`.
- Also, `throw_scan_error` is now redundant. You should remove it.

enjoyyourcopepods: The rescue block with `fail BLAST_DATABASE_ERROR.new(cmd, err)` won't work, I think. When `sys` raises an exception nothing will be returned to `err`, therefore `err` cannot be passed to `BLAST_DATABASE_ERROR.new(cmd, err)`.

yeban: Right. You will have to `rescue CommandFailed => e` and use `e.stderr`.

yeban:
Thanks!

PR #323, opened 2017-01-04
Added forking and 'exec' to the 'sys' method

enjoyyourcopepods:
The 'sys' method now forks. The child changes PATH, changes directory, and
executes a shell command using 'exec'. 'sys' waits for the child process to end, then checks the exit status of the child process.


yeban:
Thanks.

PR #326, opened 2017-01-06
Insert the 'sys' method into more places

enjoyyourcopepods:
Inserted the 'sys' method into more places where shell commands are run.

### DIFF
--- a/bin/sequenceserver
+++ b/bin/sequenceserver
@@ -2,6 +2,7 @@
 require 'readline'
 require 'English'
 require 'slop'
+require 'sequenceserver'
 
 ENV['RACK_ENV'] ||= 'production'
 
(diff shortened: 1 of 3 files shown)


bin/sequenceserver
```diff
@@ -2,6 +2,7 @@
 require 'readline'
 require 'English'
 require 'slop'
+require 'sequenceserver'
```
yeban: Requiring `sequenceserver` here will likely cause SequenceServer to default to development mode.

lib/sequenceserver/database.rb
```diff
@@ -168,12 +168,11 @@ def make_blast_database(file, type)
         _make_blast_database(file, type, title, taxid)
       end
 
-      def _make_blast_database(file, type, title, taxid, quiet = false)
```
yeban: Using `sys` changes the behaviour of this method to `quiet = true` by default. Here, we want the output of `cmd` to be printed to the console.

The fourth argument is provided to silence the output of this method during testing.

lib/sequenceserver.rb
```diff
@@ -289,7 +289,7 @@ def xdg?
 
     # Return `true` if the given command exists and is executable.
     def command?(command)
-      system("which #{command} > /dev/null 2>&1")
+      sys("which #{command}")
```
yeban: Changes the expected return value of `command?`.

PR #328, opened 2017-01-09
Have the 'sys' method replace the 'export_bin_dir' method

enjoyyourcopepods:
If a directory for NCBI BLAST+ is provided, the 'sys' method will set PATH to this directory before executing a shell command, if no other directory for PATH was provided.

Removed the redundant 'export_bin_dir' method that formerly set PATH to the NCBI BLAST+ directory.

### DIFF
--- a/lib/sequenceserver.rb
+++ b/lib/sequenceserver.rb
@@ -93,7 +93,8 @@ module SequenceServer
       child_pid = fork do
         # Set the PATH environment variable to the binary directory or
         # safe directory.
-        ENV['PATH'] = config[:bin] || options[:path] || ENV['PATH']
+        ENV['PATH'] = config[:bin] if config[:bin] && !options[:path]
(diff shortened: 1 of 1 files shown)


lib/sequenceserver.rb
```diff
@@ -91,7 +91,9 @@ def sys(command, options = {})
 
       # Fork.
       child_pid = fork do
-        # Set the PATH environment variable to the safe directory.
+        # Set the PATH environment variable to the binary directory or
+        # safe directory.
+        ENV['PATH'] = config[:bin] if config[:bin] && !options[:path]
```
yeban: How about: `ENV['PATH'] = config[:bin] || options[:path] || ENV['PATH']`?

enjoyyourcopepods: Does options[:path] need to override config[:bin]?

yeban:
Thanks. I didn't take the last two commits.

PR #330, opened 2017-01-16
Handle errors with class ENOENT rather than subclasses of ENOENT

enjoyyourcopepods:
The class ENOENT in lib/sequenceserver/exceptions.rb is the superclass of classes BIN_DIR_NOT_FOUND; DATABASE_DIR_NOT_FOUND; EXTENSION_FILE_NOT_FOUND. Handle errors with class ENOENT rather than the subclasses of ENOENT.

### DIFF
--- a/bin/sequenceserver
+++ b/bin/sequenceserver
@@ -188,19 +188,6 @@ BANNER
 
         exit!
 
-      rescue SequenceServer::EXTENSION_FILE_NOT_FOUND => e
-
-        puts e
-
-        unless require?
-          puts 'You can set the correct value by running:'
-          puts
-          puts '  sequenceserver -s -r <value>'
-          puts
-        end
-
-        exit!
-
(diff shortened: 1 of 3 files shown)


bin/sequenceserver
```diff
@@ -169,30 +169,10 @@ BANNER
...
-        puts e
-
-        unless database_dir?
-          puts 'You can set the correct value by running:'
-          puts
-          puts '  sequenceserver -s -d <value>'
-          puts
-        end
+      rescue SequenceServer::ENOENT => e
+        puts e.to_s, ""
```
yeban: Doesn't `puts e` just work?

yeban:
Looking good. Thanks. You will need to update the test suite, though. Use `rake spec exclude=spec/capybara_spec.rb` to test locally. And the merge commit needs to go.

yeban:
Thanks.

PR #332, opened 2017-02-09
query params for /searchdata.json

rjchallis:
I've added support for passing in query params like `query=seqid1,seqid2` via `searchdata.json` to prepopulate the textarea with sequences from the blast databases.

params can also be specified as `query=seqid1:start-stop,seqid2:start-stop` but for now the location information is just ignored.

### DIFF
--- a/lib/sequenceserver/routes.rb
+++ b/lib/sequenceserver/routes.rb
@@ -64,37 +64,44 @@ module SequenceServer
     # Returns data that is used to render the search form client side. These
     # include available databases and user-defined search options.
     get '/searchdata.json' do
-      sequences = ['']
+      searchdata = {
+        database: Database.all,
+        options: SequenceServer.config[:options]
+      }
       if params[:query]
         # if the url contains a query param (query=seqid1:start-stop,seqid2:start-stop)
         # parse the sequence ids and retrieve the sequences to be added to the textarea
         # by updating query state in search.js.
         #
-        # Have to check both database types as we don't know what type of id we have
-        # but if one protein match is found we don't look for nucleotide matches.
+        # Have to check both database types as we don't know what type of id we have and
+        # blastdbcmd throws an error if given a set of databases with mixed type.
         #
-        # Location parameters in the query string are currently just ignored.
-        sequences = params[:query].split(',')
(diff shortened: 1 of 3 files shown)


yeban:
Thanks. I don't quite understand how will you hit the rescue block and search nucleotide databases. I don't think `Retriever` raises an error if sequences weren't found. And it's rather convoluted to use exception handling mechanism for this. If you have unique ids (which you should ensure for your site), you don't have to search nucleotide and protein databases separately. In case both nucleotide as well as a protein sequences are found for given ids, the frontend will automatically alert the users and they can decide which sequence they want to keep, or remove.

I am fine that you don't handle the general case of non-unique ids, but for me to able to do that later, I will need you to put `retrieved_seqs` in L96 and format retrieved sequences to FASTA in the browser.

rjchallis:
Hi - I've moved the FASTA formatting over to `search.js` and tried to make a few more improvements along the way.

Just to explain a couple of my choices that I've kept from the original commit: I've separated the protein and nucleotide database searches as `blastdbcmd` throws an error for the first non nucleotide database in a mixed list, e.g. `BLAST Database error: No alias or index file found for nucleotide database [/dbs/Amyelois_transitella_v1_-_proteins.fa] in search path [/sequenceserver::]`; and I've used exception handling as `blastdbcmd` throws an error if any one of the sequence ids in the list can't be found, which propagates up as a `SequenceServer::CommandFailed` error.

This meant that no sequences were returned if any one of the ids couldn't be found so this time I've moved `Sequence::Retriever` into a loop to run separately on each provided identifier. This makes even longer code in `routes.rb` but dealing with one sequence at a time does mean that I've been able to implement the retrieval of a specific range.  

And sorry for any lack of elegance - trying to add features that are useful to Lepbase to SequenceServer is pretty much my only experience of ruby so I'm sure some of my approaches could be improved.

yeban:
Dear @rjchallis 

I am sorry for not having merged this PR. There were a few aspects of the implementation that I didn't like, and back then I didn't have the foresight either to advice how to best go about it.

I have implemented this feature in a different way in #382. I would really appreciate your comments on the PR

Priyam

Issue #338, opened 2017-09-19
minor website bugs


realsobek:
I found some minor bugs on SequenceServer's website.
Because I could not find a repo for the website I file the bug report here.

* http://www.sequenceserver.com
The navigation bar overlaps to top part of the content.
You have to delete the navigation bar (f.e. using HTML developer tools; in Mozilla Firefox: Ctrl+Shift+C) in order to see the content below it.
In the HTML-code this is the starting line of the navigation bar:
````
<nav class="navbar navbar-fixed-top" role="navigation">
````

All the following bugs are spelling mistakes.
After the website's URL the first line displays the current spelling. After the "--" separator I propose a fixed version.

* http://www.sequenceserver.com/doc/#basics
preecendence
--
precendence

* http://www.sequenceserver.com/doc/#database
````
The BLAST search algorithms don't directly understand FASTA files.
--
The BLAST search algorithms doesn't understand FASTA files.
````

````
BLAST includes the makeblastdb tool that is used to converts FASTA files
--
BLAST includes the makeblastdb tool that is used to convert FASTA files
````

* http://www.sequenceserver.com/doc/#ncbi-databases
datbases
--
databases

* http://www.sequenceserver.com/doc/#plugin
Sequenceserver provides ...
--
SequenceServer provides ...

* http://www.sequenceserver.com/doc/#sharing
Administrator previlege
--
Administrator privilege

* http://www.sequenceserver.com/doc/#macosx
Autostart on Mac OSX
--
Autostart on Mac OS X

* http://www.sequenceserver.com/doc/#issues
Ths appears
--
This appears

* http://www.sequenceserver.com/doc/#blast
commerical
--
commercial

* http://www.sequenceserver.com
````
Useability
--
Usability
````

````
preferrably
--
preferably
````

````
Envrionmental
--
Environmental
````

````
Comparitive
--
Comparative
````

````
Tuebigen
--
Tuebingen
better: Tübingen
HTML-code of ü: &uuml;
````

````
(eg:
--
(e.g.:
````

yannickwurm:
Dear @realsobek, 
thank you so much for highlighting these issues. We will rapidly address them. 
FWIW, I believe the website is served directly from the `gh-pages` branch of this repo. 
Kind regards,

Yannick


---------------------------------------------------------------------------------
https://wurmlab.github.io    y.wurm@qmul.ac.uk    Ants, Bees, Genomes & Evolution
skype:yannickwurm +44 207 882 3049
[...]

PR #339, opened 2017-09-20
fix minor website bugs #338

realsobek:
-fix spelling mistakes
-stop header navigation bar from overlapping content

Signed-off-by: René Genz <liebundartig@freenet.de>

### DIFF
--- a/css/seqserv.css
+++ b/css/seqserv.css
@@ -3,8 +3,6 @@ body {
   font-family: "Source Sans Pro","Helvetica Neue",Helvetica,Arial,sans-serif;
   position: relative;
   padding-top: 70px;
-  letter-spacing: .02em;
-  word-spacing: .01em;
   font-size: 16px !important;
 }
 
--- a/doc/index.html
+++ b/doc/index.html
@@ -57,6 +57,13 @@ layout: page
         <li
           role="presentation">
           <a
+            href="#autostart-systemd">
+            Autostart with systemd
+          </a>
+        </li>
+        <li
+          role="presentation">
+          <a
             href="#ubuntu">
             Autostart on Ubuntu / Bio Linux
           </a>
@@ -65,7 +72,7 @@ layout: page
           role="presentation">
           <a
             href="#macosx">
-            Autostart on Mac OSX
+            Autostart on Mac OS X
           </a>
         </li>
         <li
@@ -180,7 +187,7 @@ sequenceserver -s -b /path/to/latest/blast/binaries
       <p>
         The following table lists all configuration values accepted by
         SequenceServer through the configuration file or through command line
-        options. Command line options take preecendence over the values in
+        options. Command line options take precendence over the values in
         configuration file.
       </p>
       <table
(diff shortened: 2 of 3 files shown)


doc/index.html
```diff
@@ -372,9 +372,9 @@
         </h2>
       </div>
       <p>
-        The BLAST search algorithms don't directly understand FASTA files.
+        The BLAST search algorithms doesn't understand FASTA files.
```
yeban: "do" goes with plural and "does" with singular, so the original was correct

css/seqserv.css
```diff
@@ -9,7 +9,7 @@ body {
 }
 
 section {
-  padding-top: 50px;
```
yeban: First, I am unable to reproduce the navbar issue you describe in #338 in Safari, Chrome, or Firefox. Second, this will increase vertical spacing throughout the page as each section of the page is encolsed in a `<section>` tag. If what you say is indeed an issue, top padding will need to be increased specifically for the first section of the page.

doc/index.html
realsobek: indeed, my fault. I will remove the change with a rebased re-commit

css/seqserv.css
realsobek: You can reproduce the problem with this setting:
firefox -- about:preferences#content -- Fonts & Colors -- [Advanced...] -- [] Allow pages to choose their own fonts, instead of my selections above

This expands the navbar from one to two rows.
Because the content does not get pushed further down automatically the top navbar with two rows overlaps the content.

Possible solutions:
1. remove font size
# bad effect: font gets a little bit smaller; user can scale the content, if required
seqserv.css:2
delete: font-size: 16px !important;

2. remove width-constraints of top navbar
# bad effect: two right-most navbar items might get hidden, depending on resolution

bootstrap-3.3.6.min.css:14
disable: width: 1170px; 970px; 750px;

3. move start of content further down
# bad effect: target link, f.e. http://www.sequenceserver.com/#team, is covered by top navbar

seqserv.css:91
padding: 0 20px 20px 20px;
0 => 50px

seqserv.css:71
padding-top:10px;
10 => 50px


Is it OK to use solution 1?

yeban: The first option will result in using Bootstrap 3's default font size of 14. Any font size below the current 16 px seems to work. Alternatively, removing the `letter-` and `word-spacing` styles also work.

I think the ideal solution will be to isolate the use of `letter-` and `word-spacing` styles to when the Source Sans Pro font (the intended font) is used, as these styles were added specifically for that font. If that's not possible, I would remove `letter-` and `word-spacing` styles altogether instead of reducing the font size to 14.

realsobek:
The `line-` and `word-spacing` attributes have been removed from `css/seqserv.css`

With a previous recommit the change has been extended by:
-fix links to Guy Leonard

I added this change with the latest commit (sorry for feature creep)
-add paragraph: systemd autostart

yeban:
Thanks! Which distros are using systemd? Should the title be more like "Autostart on Ubuntu 16.xx (systemd)"?

realsobek:
Most of them, see:  
https://en.wikipedia.org/wiki/Systemd#Availability  
The same table show that systemd was available since Ubuntu 13.04 and is the default init system for Ubuntu since version 15.04.
On some versions you can use both, Upstart and systemd.

According to:  
https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions  
the supported versions of Ubuntu at the moment are:
14.04 LTS (supported until April 2019)
16.04 LTS
17.04

From my personal experience the default values are seldomly changed.
Hence, I would rather not include a specific version number.

Issue #201, opened 2015-09-16
Maybe use sprockets for asset management.


yeban:
With the aim of providing a way to extend and override the views and js and css.

PR #347, opened 2017-12-14
Updated to sinatra 2

homonecloco:
Hi, 
I made this minor update to update to Sinatra 2.0, since I'm integrating sequence server with a Rails 5, and 1.4 is not supported for Rails 5. It seems to be working. 
Best, 
Ricardo.

### DIFF
--- a/sequenceserver.gemspec
+++ b/sequenceserver.gemspec
@@ -17,12 +17,12 @@ interface for use locally or over the web.
 DESC
 
   # dependencies
-  s.add_dependency('sinatra',   '~> 1.4',  '>= 1.4.5')
+  s.add_dependency('sinatra',   '~> 2.0',  '>= 2.0')
   s.add_dependency('json_pure', '~> 1.8',  '>= 1.8.2')
   s.add_dependency('ox',        '~> 2.1',  '>= 2.1.1')
   s.add_dependency('slop',      '~> 3.6',  '>= 3.6.0')
 
-  s.add_development_dependency('rack-test',       '~> 0.6',  '>= 0.6.2')
+  s.add_development_dependency('rack-test',       '~> 0.8',  '>= 0.8.2')
   s.add_development_dependency('rspec',           '~> 2.8',  '>= 2.8.0')
   s.add_development_dependency('rake',            '~> 10.3', '>= 10.3.2')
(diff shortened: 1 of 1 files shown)


yeban:
Thanks for the pull-request. In my understanding, upgrading to Sinatra 2.0 will also constrain us to Ruby 2.2+. We cannot do that on the 1.0.x branch as it is meant to run on Ruby 1.8.7+.

homonecloco:
OK, would it be ok if I change the code so it decides depending on the version of Ruby?

yeban:
I am not sure that's possible. In my understanding, conditionals in the gemspec file are evaluated when the gem is built instead of when the gem is installed, which doesn't solve our problem.

For your use case, you can install your modifications to SequenceServer directly from GitHub using bundler:

    gem 'sequenceserver', git: 'https://github.com/homonecloco/sequenceserver', branch: 'update-sinatra'

hryk:
@yeban I would like to use sequenceserver with rack 2.0 too. Would it be ok for you that I create pull request for master branch?

yeban:
Yes, would be great if this PR were made for the master branch.

hryk:
@yeban I've created PR #352 .

yeban:
Closed by #352

yeban:
@homonecloco @hryk Do you know how easy or hard it would be to integrate SequenceServer with Sprockets, while still being able to use ES6, JSX and React? We use SystemJS module loader along with jspm for package management. I am okay for this to change (if need be).

homonecloco:
Hi @yeban , 
I didn't try to integrate sequence server with sprockets, the way I get to integrate it to my application is by letting it be in its own page, and trapping it in an ```<iframe>```. Then I adapt the results from the results. You can see it here: http://www.wheat-tilling.com, and in here https://github.com/homonecloco/expvip-web/blob/f404efbfb6fa05a4a8d93d69a87cd9edd261101c/app/assets/javascripts/application.js#L102

hryk:
@yeban I have't try integrate SequenceServer with my application because it could not be integrated with rails 5. So I don't know how hard it would be to integrate SequenceServer with ES6, JSX etc. I will try integration and report if integration is too hard.

yeban:
@homonecloco That's a very cool site you have created!

yeban:
@hryk Thanks! I will really appreciate that. Maybe we can move the discussion to #201

PR #363, opened 2018-05-24
circos error message updated when no hits found

aniarya82:
Previously when no hits are found the Circos error messages was not appropriate.
screen shot 2018-05-24 at 16 40 07
Therefore I added a condition to make it look similar to others.
screen shot 2018-05-24 at 16 40 50


yeban:
Thanks!

PR #365, opened 2018-05-29
added toggle button for database selection

aniarya82:
screen shot 2018-05-29 at 15 53 53
Note: Currently if a single database is selected and then the user clicks toggle all, then the selected database gets unselected and the ones not selected gets selected

### DIFF
--- a/public/js/search.js
+++ b/public/js/search.js
@@ -491,21 +491,20 @@ var Databases = React.createClass({
         this.setState({type: type});
     },
 
-    handleToggle: function (type) {       
-        var len = $('input[name="databases[]"]').length;
-        for(var i = 0; i < len; i++) {
-            var box = $('input[name="databases[]"]')[i];
-            if(box.attributes[3].nodeValue != type) {
-                box.checked = false;
-                continue;
-            };
-            if(box.checked) {
-                box.checked = false;
-            } else {
-                box.checked = true;
-            }
+    handleToggle: function (type) {
+        $(` .${type} .database `).click();
+        var elem = $(` .${type} `);
+        elem[0].innerHTML = (elem[0].innerHTML === "[Deselect All]") ? "[Select All]" : "[Deselect All]";
+        var othr = (type === "nucleotide") ? "protein" : "nucleotide";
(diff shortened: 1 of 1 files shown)


public/js/search.js
```diff
@@ -491,6 +491,24 @@ var Databases = React.createClass({
         this.setState({type: type});
     },
 
+    handleToggle: function (type) {
```
yeban: A much simpler way exists:
```
$(`.${type} .database`).click();
```

public/js/search.js
```diff
@@ -504,7 +522,18 @@ var Databases = React.createClass({
                                     className="panel panel-default">
                                     <div
                                         className="panel-heading">
-                                        <h4>{category[0].toUpperCase() + category.substring(1).toLowerCase() + " databases"}</h4>
+                                        <h4 style={{display: "inline"}}>{category[0].toUpperCase() + category.substring(1).toLowerCase() + " databases"}</h4>
+                                        <p className="" style={{display: "inline"}}>
```
yeban: Sorry, I think left aligned with the header will be better:

screen shot 2018-05-30 at 12 08 31

```
<div
    className="panel-heading">
    <h4 style={{display: "inline"}}>{category[0].toUpperCase() + category.substring(1).toLowerCase() + " databases"}</h4>
    &nbsp;&nbsp;
    <a disabled={true}
        style={{cursor: "pointer"}}
        onClick={
        _.bind(function () {
        this.handleToggle(category)
        }, this)
        }>
        [Toggle all]
    </a>
</div>
```

public/js/search.js
yeban: Further,

1. When nucleotide databases are selected, 'Toggle all' button for protein databases should be disabed and vice versa.
2. 'Toggle all' button should only be shown when there are two or more databases.

yannickwurm:
I'll add that many of us don't know what "toggle" means. So best to use Select and Deselect All 

( I had a moment of doubt, we want [deselect not unselect](https://english.stackexchange.com/questions/18465/unselect-or-deselect#18466) )

PR #382, opened 2018-10-12
Ability to pre-populate textarea with server-sent query sequences

yeban:
API is the same as proposed in PR #332. Although our implementation differs considerably, code and ideas were borrowed from the original PR.

To test this feature, run SequenceServer in development mode (`-D`). The example below uses the sample database included with SequenceServer.

```
# git clone, fetch, etc. Then,
cd sequenceserver
bundle exec bin/sequenceserver -D -d spec/database/sample
```

Now visit http://localhost:4567/?query=SI2.2.0_06267,foo,SI2.2.0_13722:1-10,SI2.2.0_80257:a-b.

Each database is searched for each locus. The search is stopped once a hit is found. In future, it may be worth memoizing the database in which a given locus was found.

/cc @rjchallis

### DIFF
--- a/lib/sequenceserver/database.rb
+++ b/lib/sequenceserver/database.rb
@@ -36,6 +36,19 @@ module SequenceServer
 
     attr_reader :id
 
+    def [](accession, coords = nil)
+      cmd = "blastdbcmd -db #{name} -entry '#{accession}'"
+      if coords
+        cmd << " -range #{coords}"
+      end
+      out, = sys(cmd, path: config[:bin])
+      out.chomp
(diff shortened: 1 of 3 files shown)


rjchallis:
Finally had a chance to try this out, thanks for reimplementing this.

This works just as I had hoped in my earlier PR, so will be a really useful new function to have for linking with other web services.  It's good to see that you are able to add comments to show which loci couldn't be found without tripping up BLAST.

yeban:
Thanks for reviewing and for proposing the feature in the first place!

PR #385, opened 2019-02-17
Updated links with PFAM; search also title, not only ID

tomas-pluskal:
Hi, 
I updated the default links-generating script:
- added Pfam and Rfam IDs
- search for IDs also in sequence title, not only in sequence ID
- fixed Uniprot name to UniProt
- updated http links to https


yeban:
Thanks, @tomas-pluskal

Issue #390, opened 2019-04-02
Rescan databases dir


Bjoernsen:
I have a central resource managing e.g. fasta files and building the corresponding blast+ indices automatically.

I have a fully automated deployment pipeline of blast+ indices to the sequenceserver database directory that runs often.

Everything works perfect. Except that I must restart the docker container(s) for the sequenceserver manually (without killing running jobs).

I know that you are not a fan of `Automatically detect new databases` #325

What do you think about a configuration added to `sequenceserver.conf` to enable the detection with `false` as default?

I already played around a bit and could offer a pull request.

yeban:
I am not up to date on docker. In the old scheme of things, Apache+Passenger should be able to handle restarts gracefully out of the box. That is, when you indicate to Passenger that the app needs to be restarted, it will start a new process and start directing all subsequent requests to the new process and terminate the old process once all previously queued requests have been handled. Alternatively, Puma and a couple other (unicorn?) Ruby application servers can handle restarts gracefully. I will be surprised if you can't do something similar automatically using docker compose or so.

Bjoernsen:
How the sequenceserver is hosted is not the issue. My problem is that I have a complete automated system/pipeline. A user uploads a fasta file and the system validates, creates the blast+ index and does many other jobs. One job is the 'deployment' to the sequenceserver data folder using symlinks.

Unfortunately, the sequenceserver does not 'recognize' the new blast db until it has been (manually) restarted.

Of course I could use e.g. passenger to host the application and write a `restart.txt` or similar to force a restart. But I thought it would be easier to enable the rescan via parameter on the sequenceserver side.

magelan:
I like the approach, that you can configure if a database rescan happens at every request, or just at serverstart. 
With the rescan_databases_dir function from the pull request, we could also build an endpoint to do so on demand.

yeban:
I am happy to have an option in SequenceServer to rescan databases, and it can be in the direction of #391 (which is a very clever hack, imo) provided my concern about race condition is addressed.

Bjoernsen:
How can I modify my MR #391 to get merged?

yeban:
[...]

PR #391, opened 2019-04-02
enable rescan databases dir #390

Bjoernsen:
https://github.com/wurmlab/sequenceserver/issues/390


yeban:
The `@collection` instance variable is per-process. If requests are handled by a server in different threads, it is possible that one thread clears `@collection` while another thread is trying to access it and cause a race condition.

PR #451, opened 2020-01-28
Optimise rendering process

yeban:
First update transfers responseJSON to Report.state so that summary
overview and circos can be rendered. Subsequent updates add Query, Hit,
and HSP components to Report.state.results which are then inserted into
the page by React. The first cycle is started with the call to setState
in setStateFromJSON and subsequent cycles are is queued in
componentDidUpdate. componentDidUpdate is run after the DOM has been
updated (that is, virtual DOM has been applied). The use of setTimeout
ensures that the browser gets a break from the rendering process and can
respond to user interactions. Because we queue the next update after DOM
has been updated, the setTimeout interval does not need to be very high.
Finally, the number of each HSPs rendered in a cycle should be small.
This is because the browser is blocked from responding to user
interactions such as click till it has finished the current cycle.
However, if this number is too small, like 1, the time required to
render the whole report increases. The implemented threshold of 3 is 3
times faster than rendering only 1 HSP in each update cycle.

The rendering process is further optimised by implementing the
shouldComponentUpdate life-cycle method for Query, Hit, and HSP
components. This avoids re-rendering (including to virtual-dom).
This causes the view sequence button per-hit to break as it
relies on Hit.state to activate SequenceViewer.

API wise, hits are no longer nested inside queries, and hsps are no
longer nested inside hits. This has created minor breaking changes
that should be addressed next.


yeban:
The PR breaks down the rendering of BLAST results into small slices. This fixes the problem where large reports would just fail to load because the browser would get stuck rendering the results all at once and eventually get killed itself due to excessive resource usage or by the user for being unresponsive.

The PR introduces a few code-smells as highlighted by code-climate. However, addressing them would require changing response JSON to be flatter. Without changing the JSON schema, it might be possible to refactor `updateState` using promises or callbacks, but I don't think that would reduce the cognitive overhead.

The PR works. Travis indicates a failing test on Ruby 2.5 which I can reproduce locally. However, it is only one test case that fails and it fails only on Ruby 2.5: I tested locally on Ruby 2.3 - 2.6. This suggests a problem with the dependencies that are being pulled for Ruby 2.5.

I will check if capping dependencies can fix the failing example. Otherwise, the PR is ready for testing on other PCs and to base further work on.

PR #471, opened 2020-06-09
add few likely search terms

yannickwurm:


views/layout.erb
```diff
@@ -74,6 +74,10 @@
           href='http://www.sequenceserver.com'>
           <%=
             quote = [
+	      "Run BLAST on local computer.",
```
yeban, line 77: Would be great to match indentation with the lines below.

PR #472, opened 2020-06-11
Update sinatra

homonecloco:
Hi, 
Some time ago I tried to submit a pull request to update ```sequenceserver``` to use Sinatra 2.0, as this would allow the integration on rails5 projects. Last time, I got the feedback that it would break compatibility with legacy versions of ruby. To address this, I modified the gemspec to conditionally select the version of Sinatra, depending on the ```RUBY_VERSION```. Please test that the changes worked, as I had issues installing ruby 1.8 and 1.9, but if you can't test it, and ar interested in the patch, I can invest more time testing. 

  I hope you consider the this pull request.


homonecloco:
Ok, I figured out how to test this with travis without installing everything, I'll fix the building to make sure the builds still pass across rubies.

homonecloco:
Ok, now it passes the travis tests. The automatic test failed because one of the jobs couldn't download blast. However, the rest of the tests ran correctly (and it was ok on my own Sinatra test). I also added to travis up to ruby 2.7.

yeban:
Hi. I can't install Ruby 1.8 and 1.9 either - fails to compile.

The pull-request looks alright to me though and I don't mind merging it if you absolutely need this change upstream. But have you seen the candidate release of SequenceServer 2.0? It uses Sinatra 2.0 and has a lot of other improvements. More info here: https://groups.google.com/forum/#!topic/sequenceserver/c98ePBzcuVE. Also, note that we won't be maintaining 1.0.x branch further once 2.0 release has graduated to stable.

If you still want to get the patch merged into 1.0.x branch, two thoughts on the code, in that order of importance:
1. Would be nice if the sinatra and rack-test dependencies for Ruby >2.3 were specified the exact same as in the master branch. Otherwise, 1.0 and 2.0 combined we would be working with three versions of rack-test which can be confusing later one.
2. Would be nice to use `&&` over `and`. This is just `&&` is more universal and is less of a context switch for people working simultaneously with multiple programming languages (including myself).

homonecloco:
Hi Yeban, 
Thanks for the feedback. I didn't test on 1.8 and 1.9 as I didn't see them listed in the .travis file, so I wasn't sure if they were officially supported or not. 

I had seen version 2.0. I really like what you are doing there, but for my case I will probably continue using version 1.0, as my main use case is to integrate sequence server in other websites that only require the the coordinates or the genes that result.  

I'll have a look on how the rack-test are in the master branch, I tried to use a few changes as possible in the code on the 1.0.x branch. 

Best, 
Ricardo.

PR #473, opened 2020-06-28
Gh pages2.0

richarda23:
Hi
 This PR includes a Markdown file describing how to access SequenceServer programmatically, using curl commands to exemplify. There is a also a bash script to illustrate how all the commands can be linked together to submit a BLAST job, wait for completion and download the results.

Please suggest improvements or things I've missed out on, this documentation is really just a tidied up documentation  of what I've discovered.
Thanks for your consideration, 
Richard


yeban:
This is brilliant! Thanks!

Please could you,
1. Add two `---` to the top of the `.md` file, so that it is automatically converted to `html` by github/jekyll. To test, run `bundle exec jekyll serve` and go to `http://localhost:4000/apiDoc/api`
2. Fix a few rendering issues that I am seeing on my Mac here (screenshot below).

Screenshot 2020-06-29 at 14 23 34

If you don't mind, I would prefer the document to be accessible at `/doc/api`. This can be done in one of the following two ways, whichever you prefer:
1. Simply move both the files (`api.md` and the script) to `doc/`, or
2. Move `apiDoc` to `doc/api/` and rename `api.md` to `index.md`

Finally, a way to make the document accessible to users would be to add a section for API in `doc/index.html` which simply links to `api.md`. You are welcome to do this if you like, but I wouldn't mind doing it myself - your document is already a big step forwards.

yeban:
Thanks!

I fixed the content-type issue you reported - will be available in next release.

PR #487, opened 2020-10-03
Scan and update databases on page reload

alisahari9:
Currently, when we add a new database, we must kill the server and start it again so the change shows up on the page. This means a downtime for the server. Here, I suggest adding a line to scan the databases on every page reload and eliminate the need to kill the sever.

### DIFF
--- a/lib/sequenceserver/blast/job.rb
+++ b/lib/sequenceserver/blast/job.rb
@@ -36,7 +36,7 @@ module SequenceServer
       def raise!
         return true if exitstatus == 0
 
-        stderr = File.readlines(this.stderr)
+        stderr = File.readlines(self.stderr)
         if exitstatus == 1 # error in query sequence or options; see [1]
           error = stderr.grep(ERROR_LINE)
           error = stderr if error.empty?
--- a/lib/sequenceserver/routes.rb
+++ b/lib/sequenceserver/routes.rb
@@ -56,6 +56,7 @@ module SequenceServer
 
     # Returns base HTML. Rest happens client-side: rendering the search form.
     get '/' do
+      Database.scan_databases_dir
       erb :search, layout: true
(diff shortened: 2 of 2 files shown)


Bjoernsen:
See https://github.com/wurmlab/sequenceserver/issues/390 and https://github.com/wurmlab/sequenceserver/pull/391

The problem is the number of databases. If you have a few databases, it works as you suggested. But if you have many BLAST DBs, it will take some while (> 30s) for each request to get a list of all databases and to load the page. I promise that you colleagues will complain 😏

alisahari9:
I see, good point. Since this change would not be ideal in case we have many BLAST databases, please feel free to close this PR.

yannickwurm:
Thank you both. 
I think it could work to have this as an optional flag given that both usage cases are relevant...

alisahari9:
> Thank you both. I think it could work to have this as an optional flag given that both usage cases are relevant...

alisahari9:
I accidentally closed the PR! Sorry about that! Reopened it!

yeban:
Hmm. I am not so sure of the change. It won't work if a database was removed from database directory. Same if a database was renamed. The old entries will appear in the search form. Even if such a situation arises 1% of the time, it needs to be accounted for in my opinion. We will end up looking silly otherwise. 

I have mixed feelings as well about making rescanning configurable/optional. When there are a large number of databases, the admin of the installation is unlikely to be the (only) end user. All end users of that installation will not know that a certain special flag has been activated, and will perceive the noticeable delay in loading search form as SequenceServer sucks.

Rescanning can happen in a background thread. We have example of this in the code base - the `JobRemover` class. Database deletions and renaming can be handled by creating a new collection of databases on each scan, which is then swapped with the global collection.

PR #490, opened 2020-10-14
Optimize Dockerfile by removing unused packages

### LINKED ISSUE #413: Optimise docker image
The docker image for beta is pretty big: ~500 Mb

1. ~~Add `rm -rf /var/lib/apt/lists/*` to RUN directive on line number 8 (run `git log -p Dockerfile` to see how it was done before and how it accidentally got removed).~~
2. ~~Combine RUN directive in line number 21 and 22 into one (`gem install bundler && bundle install --without=development && echo "\n/db\nn" | bundle exec bin/sequenceserver -s` - i.e., yes to download blast, /db for location of database dir, and no for scanning database dir as the database dir would be empty while building the image on docker hub)~~

3. Remove the BLAST tarball downloaded by sequenceserver from /tmp

neelkamath:
Fixes #413 

Demo:
Screen Shot 2020-10-14 at 8 53 01 PM

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,4 +22,25 @@ RUN gem install bundler && \
         yes '' | bundle exec bin/sequenceserver -s -d spec/database/sample
 RUN touch ~/.sequenceserver/asked_to_join
 
+RUN rm /root/.sequenceserver/ncbi-blast-2.10.0+-x64-linux.tar.gz
+WORKDIR '/root/.sequenceserver/ncbi-blast-2.10.0+/bin/'
+RUN rm \
+    deltablast \
+    legacy_blast.pl \
+    makeprofiledb \
+    rpstblastn \
+    blastdb_aliastool \
+    cleanup-blastdb-volumes.py \
+    dustmasker \
+    psiblast \
+    segmasker \
+    update_blastdb.pl \
+    blastdbcheck \
+    convert2blastmask \
+    get_species_taxids.sh \
+    makembindex \
+    rpsblast \
+    windowmasker
+WORKDIR /sequenceserver
+
 CMD ["bundle", "exec", "bin/sequenceserver", "-d", "/db"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,26 +19,26 @@ WORKDIR /sequenceserver
 # of starting up, while -d is used to suppress questions about database dir.
 RUN gem install bundler && \
         bundle install --without=development && \
-        yes '' | bundle exec bin/sequenceserver -s -d spec/database/sample
-RUN touch ~/.sequenceserver/asked_to_join && \
-    rm /root/.sequenceserver/ncbi-blast-2.10.0+-x64-linux.tar.gz && \
-    cd '/root/.sequenceserver/ncbi-blast-2.10.0+/bin/' && \
-    rm \
-        deltablast \
-        legacy_blast.pl \
-        makeprofiledb \
-        rpstblastn \
-        blastdb_aliastool \
-        cleanup-blastdb-volumes.py \
-        dustmasker \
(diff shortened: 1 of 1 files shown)


yeban:
Thanks. I build the updated Dockerfile my Mac here. Interestingly, the image size has gone up from 1.42 to 1.58 GB:

Screenshot 2020-10-14 at 16 26 42

Could this be because the new RUN directives are creating a new layer each? The increase in image size defeats the purpose of deleting unused files.

neelkamath:
Interesting. I get a smaller image size on my Mac. I don't know if this is because of BuildKit or something. I'll reduce the steps nevertheless.

neelkamath:
Pulling from Docker Hub, I get 1.42 GB, but building on `master` branch I get 1.47. On my branch I also get 1.47. Can you check?

yeban:
There is a similar size discrepancy on my end as well: pulling from docker hub I get 1.42 Gb, building on master branch I get 1.58 Gb, building from your branch I get 1.58 Gb.

The size discrepancy between docker hub and local build is intriguing, but if we keep that aside for a moment, image size should be lower on your branch than on master. I think this is not the case because you remove files in another layer. To keep image size small, I think it is important to remove intermediate files in the same layer that they are added in. Which is why in most Dockerfiles you will see `rm -rf /var/lib/apt/lists/*` in the same line as `apt update` and `apt install`.

neelkamath:
Since there were two `RUN` layers on the `master` branch, I merged them into one along with my own changes. The image is now 866MB instead of 1.47GB on my machine. I'm unsure why there were two `RUN` commands grouped together, so please verify the changes work as expected.

yeban:
That's ~40% reduction in image size. Thanks a lot!

yannickwurm:
THanks
Wouldn't moving line 7 to line 20 and combining those have a similar effect?

Yannick

On 2020Oct15, at 16:55, Anurag Priyam <notifications@github.com<mailto:notifications@github.com>> wrote:



That's ~40% reduction in image size. Thanks a lot!

yeban:
@yannickwurm That will reduce one layer in the image, but it won't make a big impact with respect to size. It is the deletion of unused files (at the right point during image construction) that made a big impact here.

Further scope for optimising image size, in my opinion, is as follows:
1. Selectively install required packages instead of installing the build-essential meta package. This will require some experimentation to figure out what we need and what we don't.
2. By getting rid of `git ls-files` in gemspec, we can avoid installing git as well in the container.
3. spec/ is 79 Mb on my system. One could argue it is not relevant to include it. I don't know if COPY respects .dockerignore. If yes, maybe we can just ignore it there.
4. An idea I have considered in the past but not very sure of the implementation and the impact it will have on user-experience is to not bake BLAST binaries in the image. Instead, download them on the first run (like a typical non-Docker setup) to a persistent volume. Such an approach is taken by, for e.g., mysql where the database folder (`/var/mysql` or something) is created on the first run in a persistent docker volume.

neelkamath:
@yeban yes, the `COPY` command respects the `.dockerignore` file. Usually a `.dockerignore` would look like:

```ignore
*
!src/
!package.json
```

and then the `Dockerfile` would simply say `COPY . .`.

yeban:
@neelkamath Do you know of a way to ignore everything in `spec/` except `spec/database/sample`? The `spec/database/sample` directory is used in line 22 of Dockerfile, so we need to keep that.

yannickwurm:
Avoid 4. Because it would hurt instanciation time
________________________________

neelkamath:
@yeban I believe it'd be similar to other ignore files, probably something like:

```ignore
spec/
!spec/database/sample
```

Issue #491, opened 2020-10-16
Pre-fill sequence textarea on default page with POST (without databases[] or advanced)


sammyjava:
Pardon me if this is a repeat, but I've not found it anywhere. I'd like to simply send a sequence to the SequenceServer via GET or more likely POST and have the user then choose the databases, add any BLAST options, and submit the search. I've been able to do the full POST of method+sequence+databases[]+advanced, but it seems a bit silly for me to rebuild that form when all I really want to do is get a sequence from my webapp (InterMine) over to SequenceServer. Is there already a way to do this? Thanks!

sammyjava:
A great way to implement this would be to make databases[] not required by the POST API (sequence and method could remain required, or maybe not even method). If databases[] is not supplied, display the search form with the sequence (and method) set from the POST vars.

yeban:
If the query sequence is contained in the BLAST databases, you can pre-fill the sequence textarea using its id:

`/?query=sequence_id`
`/?query=sequence_id:start_cord-stop_coord`

If the above doesn't work for you, pre-filling on POST could also be implemented, but I would be more in the favour of using Javascript to open sequenceserver in a new window and paste in the query sequence. It is 2020, surely the latter should be possible in a cross-browser manner.

sammyjava:
The query sequence is arbitrary. We're blasting it against the BLAST databases to see what we get. It's FASTA text.

I agree that Javascript could be used to launch a new window with the texteara pre-filled, but that requires that anyone that wishes this functionality needs to code it themselves, while a simple presentation of the search page with the textarea prefilled from a sequence POST works "out of the box" for the world. Sometimes server-side solutions are preferable, even in 2020.

yeban:
I am a bit confused about what you mean by "out of the box". On the page that links to SequenceServer, you would implement a form that makes the post request, right?

Regarding the Javascript solution what I had in mind was to provide a source file that one could load on the page and activate with a URL, and HTML id of two elements, one to read the sequence from and one for the 'submit/blast' button. Something like:

```html
<script src="/sequenceserver.js"></script>
<script>
SequenceServer.sendTo('url', 'id_of_sequence_field', 'id_of_blast_button')
<script>
```

sammyjava:
> I am a bit confused about what you mean by "out of the box". On the page that links to SequenceServer, you would implement a form that makes the post request, right?

[...]

PR #493, opened 2020-10-28
Update routes.rb

sammyjava:
Switched `erb : layout` to `erb :search, layout: true` under `if params[:input_sequence]` so that (1) an error isn't thrown and (2) one gets a pre-filled sequence textarea if one posts the variable `input_sequence` from elsewhere. I believe this was a bug, since posting `input_sequence` generated an error. This solves issue #491 .


yeban:
Thanks!

Issue #488, opened 2020-10-06
Improve IP/hostname detection on EC2


yeban:
SequenceServer shows private IP and URL of an EC2 instance on startup. I wonder if it is possible for it to be more accurate here.

yeban:
AWS free tier can be used to debug this

tuxinaut:
> I wonder if it is possible for it to be more accurate here.

Would you mind to elaborate a bit further on this?

yeban:
SequenceServer tries to detect the IP of the computer it is running on and constructs a URL from it for sharing, like "to share the setup with your colleagues, try using http://abc.def.ghi.jkl:4567". For this to be helpful, the IP SequenceServer detects should be the public IP.

AWS EC2 instances have both a public and private IP. Public IP is the one we use to ssh to the instance, etc. When running SequenceServer on an EC2 instance, we would like SequenceServer to detect the public IP and use that for showing the sharing URL, but it only detects the private IP.

Does that help?

tuxinaut:
> Does that help?

Thanks, it makes no more sense.

yeban:
Well, the issue refers to a message displayed by SequenceServer on startup:

```
** SequenceServer is ready.
   Go to http://localhost:4570 in your browser and start BLASTing!
   To share your setup, try one of the following addresses. These
   may only work within your home, office, or university network.
     -  http://192.168.1.104:4570
```

We want the URL in the line `http://192.168.1.104:4570` to use the public IP of the computer. IP detection code is here: https://github.com/wurmlab/sequenceserver/blob/master/lib/sequenceserver.rb#L253. I suppose if it is not possible to differentiate between public and private IP, we can display both.

tuxinaut:
I'll give it a try :)

PR #494, opened 2020-10-30
Improve IP/hostname detection on EC2 (first try)

tuxinaut:
It's not the best code on planet earth but it does what it's supposed to do.

I would plan to improve the code after this PR and even though Hacktober is over ;)

Related to #488

### DIFF
--- a/lib/sequenceserver.rb
+++ b/lib/sequenceserver.rb
@@ -1,4 +1,5 @@
 require 'English'
+require 'net/http'
 require 'socket'
 require 'resolv'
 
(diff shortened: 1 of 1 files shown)


yeban:
Thanks. Can you help me understand the rationale of your approach? Did you consider the possibility that (in line 254) the public IP may be included in the array returned by `Socket.ip_address_list`, but is not being returned by `.find` because:
1. both public and private IP will match the condition defined within `{}` and
2. if there are multiple matches, `find` returns the first match, and the first match just happens to be the private ip

tuxinaut:
> Thanks. Can you help me understand the rationale of your approach? 

Yes, sure I will try my best. At first, I'm a bit biased regarding the AWS EC2 metadata service ;)

>  Did you consider the possibility that (in line 254) the public IP may be included in the array returned by Socket.ip_address_list, but is not being returned by .find because:

>  1.  both public and private IP will match the condition defined within {} and
>  2. if there are multiple matches, find returns the first match, and the first match just happens to be the private ip

Let me (re-)check this possibility.

yeban:
Although I discovered the issue while working on AWS EC2, I think the problem is of a more general nature. Other cloud providers may also have a private and a public IP for their instances, and SequenceServer might fail to detect the right one. Thus a more general solution would be more preferable.

I think you will find that `Socket.ip_address_list` already includes the public ip (no, I have not tested that to be 100% sure). I also came across a method in Ruby's Addrinfo class that can help distinguish private IP: https://docs.ruby-lang.org/en/2.1.0/Addrinfo.html#method-i-ipv4_private-3F.

tuxinaut:
I will try to have another look this weekend.

yeban:
I will close this for now.

PR #496, opened 2020-11-02
Make textarea vertically resizable

Iain-S:
- Make notifications narrower so that they don't obscure the resize handle
- See issue 401


yeban:
Don't mind the bots. Happy to merge once you have had the chance to read and agree to the CLA.

Issue #456, opened 2020-03-14
setup script should elegantly handle incorrect database dir


yannickwurm:
Currently it dies - it should just ask again

```
[2020-03-14 11:11:41] INFO  Configuration file not found: /Users/Yannick/.sequenceserver.conf

Database dir not set.

SequenceServer needs to know where your FASTA files or BLAST+ databases are.
Please enter the path to the relevant directory (default: current directory).

Press Ctrl+C to quit.

>> ~/db
[2020-03-14 11:21:27] INFO  Configuration file not found: /Users/Yannick/.sequenceserver.conf
Could not find database dir: /Users/Yannick/db
➜  Developer ls ~/db
ls: /Users/Yannick/db: No such file or directory
➜  Developer ls ~/db
```

yeban:
I think the setup script should do the same for 'path to blast binaries' as well and can be done by defining a little helper method that prompts and re-prompts the user for input till valid (i.e., file exists and is a directory).

Iain-S:
I have submitted a PR for this (though I still need to get the tests passing): #499

PR #499, opened 2020-11-06
Handle incorrect BLAST+ or database dirs nicely

Iain-S:
- If a user provides incorrect directory paths during setup, ask them repeatedly for the correct ones (instead of exiting)

As mentioned in issue #456, sequenceserver exits immediately if an incorrect BLAST+ directory or database directory is given.  By raising a couple of new exceptions and catching them (as is already done for incompatible BLAST versions) we let the user try again.

### DIFF
--- a/bin/sequenceserver
+++ b/bin/sequenceserver
@@ -184,8 +184,8 @@ begin
       # error scenarios either when creating a new SequenceServer (first
       # time or later), or updating config values using -s CLI option.
       rescue SequenceServer::BLAST_NOT_INSTALLED_OR_NOT_EXECUTABLE,
-             SequenceServer::BLAST_NOT_COMPATIBLE => e
-
+             SequenceServer::BLAST_NOT_COMPATIBLE,
+             SequenceServer::BLAST_BIN_DIR_NOT_FOUND => e
         # Show original error message first.
         puts
         puts e
@@ -254,8 +254,9 @@ begin
           puts
         end
         redo
-      rescue SequenceServer::DATABASE_DIR_NOT_SET => e
-        # Show original error message.
+      rescue SequenceServer::DATABASE_DIR_NOT_SET,
+             SequenceServer::DATABASE_DIR_NOT_FOUND => e
+        # Show original error message first.
         puts
         puts e
 
--- a/lib/sequenceserver.rb
+++ b/lib/sequenceserver.rb
@@ -181,7 +181,7 @@ module SequenceServer
       if config[:bin]
         config[:bin] = File.expand_path config[:bin]
         unless File.exist?(config[:bin]) && File.directory?(config[:bin])
-          fail ENOENT.new('bin dir', config[:bin])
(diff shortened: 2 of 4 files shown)


yeban:
Thanks. The pull-request sure addresses #456 but it also changes the behaviour of the `-b` and `-d` command line options.

On master:

```bash
~/sequenceserver master*
❯ bundle exec bin/sequenceserver -d /foo
[2020-11-14 12:27:32] INFO  Reading configuration file: /Users/priyam/.sequenceserver.conf.
Could not find database dir: /foo
```

On your branch:

```bash
~/sequenceserver remotes/iain/handle_incorrect_directories* 7m 47s
❯ bundle exec bin/sequenceserver -d /foo
[2020-11-14 12:26:57] INFO  Reading configuration file: /Users/priyam/.sequenceserver.conf.

Database directory provided doesn't exist (or isn't a directory).

SequenceServer needs to know where your FASTA files or BLAST+ databases are.
Please enter the path to the relevant directory (default: current directory).

Press Ctrl+C to quit.

>>
```

At least `Could not find database dir: /foo` needs to be said before dropping to an interactive prompt. But I would rather we retain the existing behaviour for CLI switches. The reason being, the original command I typed remains incorrect and I cannot just reuse it from history without going through SequenceServer's prompt again. I also don't find it natural to be dropped to a prompt to correct a typo when I can easily press up arrow and fix the typo - just seems unnecessary.

I think a better approach to address #456 would be to check if the path provided by the user exists and ask again if incorrect. Would be nice if the bit that asks, validates, and asks again is in a method of its own. I am sorry I couldn't respond before you went ahead and updated the tests as well.

Iain-S:
Cool, no worries, I can spin things off into a separate validation method (and also make sure that I don't change the `-d` and `-b` behaviour).

yeban:
Heads up - `bin/sequenceserver` has moved forward from when you created the pull-request

yeban:
Closing as stale. Please reopen when ready.

Iain-S:
Hi,

I had a query about desired behaviour...

On the master branch, giving a non-existant directory as a `-b` or `-d` option exits back to the command line: 

```
$ bundle exec bin/sequenceserver -b /foo
Could not find bin dir: /foo
```

However, giving an incorrect directory (i.e. one the doesn't contain blast binaries or a database) drops you to the interactive prompt:

```
$ bundle exec bin/sequenceserver -b ./LICENSE
[2021-01-12 16:13:02] INFO  Reading configuration file: /home/iains/.sequenceserver.conf.

BLAST+ is either not installed, or there is a problem with the
installed version.

If you have downloaded NCBI BLAST+ already, please enter the path
below. Otherwise just press Enter and SequenceServer will download
a copy of NCBI BLAST+ for itself.

Press Ctrl+C to abort setup.

>> 
```

Presumably, we want to exit sequenceserver, back to the command line, in both cases?

yeban:
Indeed, exiting to command line would be the desirable behaviour in both the cases.

PR #504, opened 2020-11-24
Ensure customizations to CSS & JavaScript are reflected in container image

nathanweeks:
Currently, updates to public/css & public/js aren't reflected in the minified versions used in the Docker image. It would be convenient if the Dockerfile took care of generating these. I'm not familiar with best practices for jspm in Dockerfiles, but this naive attempt seemed to work, and could be at least a starting point to refine.

Alternatively, as it looks like jspm was [deprecated in June 2020](https://github.com/jspm/jspm-cli), and it's probably worth investigating a replacement.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,7 @@
 !bin
 !config.ru
 !lib
+!package.json
 !public
 !sequenceserver.gemspec
 !views
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,18 +2,7 @@
 # for the variables to be accessible in FROM instruction.
 ARG BLAST_VERSION=2.10.0
 
-## Stage 1: CSS & JS.
-FROM node:15-alpine3.12 AS node
-
-RUN apk add --no-cache git
-WORKDIR /usr/src/app
-COPY ./package.json .
-RUN npm install
-ENV PATH=${PWD}/node_modules/.bin:${PATH}
-COPY public public
-RUN npm run-script build
-
-## Stage 2: gem dependencies.
+## Stage 1: gem dependencies.
 FROM ruby:2.7-slim-buster AS builder
 
 # Copy over files required for installing gem dependencies.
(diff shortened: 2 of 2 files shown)


yeban:
It is a good pull-request, but I don't fully understand the problem or the goal. Minified JS and CSS files are tracked by git and automatically included in the docker image (and any other distribution such as `.tar.gz`, `.gem`).

I am open to moving away from jspm-cli. I think using npm to install packages and retaining jspm's systemjs to load packages can work.

nathanweeks:
The rationale would be to allow sites to customize their local CSS & JS, or to test any changes to these that may be potentially useful upstream before submitting a PR. Although TBH this probably wouldn't be done as often as, e.g., customizing links, so it might not make sense to incur the overhead of downloading node packages & building minified assets by default (especially if no such customizations were done). Consequently, I reworked the PR to make default build the same as it was previously, with the minified CSS & JS assets built only if the "minify" build target is specified (e.g., `docker build --target=minify .`). But I can understand if it's more convenient to revisit this after any planned replacement of jspm.

yeban:
That's fair and I appreciate the move to make it optional. However, a default build on my Mac (`docker build . -t seqserv-nathan`) ended up minifying assets and copying them over:

```
~/sequenceserver css-js*
❯ docker build . -t seqserv-nathan
...
> SequenceServer@2.0.0-beta build
> jspm bundle-sfx --minify --skip-source-maps js/search.js public/sequenceserver-search.min.js && jspm bundle-sfx --minify --skip-source-maps js/report.js public/sequenceserver-report.min.js && cleancss public/css/sequenceserver.css > public/css/sequenceserver.min.css

     Building the single-file sfx bundle for js/search.js...
ok   Built into public/sequenceserver-search.min.js, minified, mangled.
     Building the single-file sfx bundle for js/report.js...
ok   Built into public/sequenceserver-report.min.js, minified, mangled.
Removing intermediate container 0536a120b9bf
 ---> 1167c3632a8d
Step 42/45 : FROM final AS minify
 ---> 7cd0718e3d5a
Step 43/45 : COPY --from=node /usr/src/app/public/sequenceserver-*.min.js public/
 ---> f6810bc5007d
Step 44/45 : COPY --from=node /usr/src/app/public/css/sequenceserver.min.css public/css/
 ---> 9cc65f9eac65
Step 45/45 : FROM final
 ---> 7cd0718e3d5a
Successfully built 7cd0718e3d5a
Successfully tagged seqserv-nathan:latest
```

There are currently no plans to migrate from JSPM but if you would be interested, you are very welcome to send a pull-request :)

nathanweeks:
Ah, right---I tested with BuildKit enabled, which, unlike the legacy builder, is [smart enough to skip unused build stages](https://www.docker.com/blog/advanced-dockerfiles-faster-builds-and-smaller-images-using-buildkit-and-multistage-builds/):
```
DOCKER_BUILDKIT=1 docker build . -t seqserv-nathan
...
```
BuildKit is enabled by default on new Docker Desktop >= 2.4.0.0 installs, or after a factory reset after updating from an older version (see the [Docker Desktop 2.4.0.0 release notes](https://docs.docker.com/docker-for-mac/release-notes/#docker-desktop-community-2400)).

Regarding JSPM: I guess my preference would be to not fix it while it ain't broke.

yeban:
I have learnt a great deal from your pull-requests. Thanks!

PR #527, opened 2021-05-26
Change instructions for running all rspec tests

Iain-S:
The current instructions in master do not work for me (Ubuntu/Bash)

```
$ bundle exec rspec spec/**/*spec*
...
An error occurred while loading ./spec/**/*spec*.
...
No examples found.
```

Perhaps the command should add `--pattern` and be `bundle exec rspec --pattern spec/**/*spec*`?

However, that pattern picks up 483 tests (including the ones in `doctor_spec.disabled.rb` and, I believe, `spec/blast_versions/blast_2.2.30/import_spec_capybara_local_2.2.30.rb`) vs 271 when no pattern is explicitly given (the default, in this case, is `spec/**/*_spec.rb`).

If you let me know whether it is desirable to pick up those extra spec files, I can modify this PR (or not) as required.


Iain-S:
On further inspection, it looks like `spec/**/*spec*` might work if the "globstar" shell option is enabled (with `shopt -s globstar`). I'm not sure how widespread it is. It may still be better to use `--pattern` if rspec's handling is more consistent across platforms.

yeban:
`spec/**/*spec*` was written with zsh in mind. I agree with you that it should be more general and that the instructions are not precise.

`.travis.yaml` codifies how the tests should be run:
1. To test core functionality: `bundle exec rspec spec/*_spec.rb spec/blast_versions/blast_2.9.0/*`
2. To run what we call 'compatibility' tests: `bundle exec rspec spec/blast_versions/*/*`. Just that it is written in an expanded form `.travis.yaml` so as to run each as a parallel job.

However, it is not important to include the second command in README - only the first one is important. This is because the compatibility tests are largely redundant, I think.

Ideal would indeed be if `bundle exec rspec` ran all the "core" tests. That would require moving the two tests files out of `spec/blast_versions/blast_2.9.0/*` and into `spec` folder, and renaming them to match rspec's default behaviour.  But that's a bit more work than running `mv`.

`spec/doctor_spec.disabled.rb` can be safely removed, if you would like.

Iain-S:
I have modified the .rspec file so that the default, if you run `bundle exec rspec`, is as you describe above to test the core functionality. That should match the travis `script:` section, I think. In fact, that section could be changed from 

```yaml
script:  
    # Run main specs and import spec for BLAST 2.9.0+  
    - bundle exec rspec spec/*_spec.rb spec/blast_versions/blast_2.9.0/*
```

to just `bundle exec rspec` as well, if you like?

yeban:
I like the addition of `.rspec` file. This won't break the commands in the `job` section of `.travis.yaml` right?

Yes, changing `script` to `bundle exec rspec` would be nice. Conforming to standard expectations has several advantages.

p.s: I broke Travis with one of my recent changes. I think it's a configuration issue. I plan to look into it on Monday.

PR #531, opened 2021-06-09
Run codeclimate eslint

Iain-S:
- Update eslint config file.
- Fix formatting errors (and a couple of others).

---

- I have "fixed" the d3 error by making it a global variable. Somewhat of a hack but it seems to be what most people do.
- I have allowed unused variables in function definitions. Personally, I find it useful to see what arguments callbacks are being passed, even if I'm not using them. An alternative would be to use `/*eslint-disable no-unused-vars*/` around all of the offending lines 🤮.
- Most of the .js file changes are purely aesthetic (e.g. quotes and indentation) but `visualisation_helpers.js` and `report.js` have some minor logic changes you may want to check.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,7 +2,8 @@
     "env": {
         "browser": true,
         "commonjs": true,
-        "es6": true
+        "es6": true,
+        "jquery": true
     },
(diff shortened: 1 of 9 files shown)


yeban:
Thanks!

PR #532, opened 2021-06-09
Add stylelint engine for .css linting

Iain-S:
## Issues with CSSLint

Running CSSLint via codeclimate has several issues:

1. Most importantly, I can't get it to pay attention to .csslintrc so we have no way to ignore rules (I'm pretty sure it also ignores commandline args when run via codeclimate so we can't do `--ignore="rule1","rule2"`).
1. CSSLint isn't maintained so many of the rule violations are actually CSSLint bugs.
1. When run via codeclimate, it doesn't print the name of the broken rule, only the offending file and linenumber.
1. If you install csslint with npm, the default formatting still doesn't print the rule name. I had to do `csslint --format=compact public/css/ |  sed 's/line/\n\nline/g'` to get a useful output. The standalone `csslint` does at least adhere to .csslintrc.

All in all, I'm not sure the codeclimate csslint plugin is that useful as a CI tool. Though you may still want to use it to find the * and *= issues (though even this is[ disputed](https://github.com/Semantic-Org/Semantic-UI/issues/5875)).

## stylelint

[Stylelint](https://stylelint.io/user-guide/rules/about) doesn't pick up all of the same issues as CSSLint but

1. The VSCode plugin is nice.
1. The standalone stylelint comandline tool has a `--fix` option to automatically fix some issues.
1.  When run with `codeclimate -e`, stylelint does print the name of the broken rule(s).

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -14,6 +14,8 @@ engines:
     enabled: true
   csslint:
     enabled: true
+  stylelint:
+    enabled: true
 ratings:
   paths:
   - "**.rb"
@@ -29,3 +31,4 @@ exclude_paths:
 - public/shims/**/*
 - public/vendor/*
 - spec/**/*
+- coverage/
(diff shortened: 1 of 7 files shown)


yeban:
Thanks!

PR #542, opened 2021-08-23
merging previous PRs into fork

discoform:


yeban:
Thanks. I think it would have been good to summarise the changes of the previous PRs here.

The website now takes more horizontal space, which is good - is that because of the bootstrap update?

On the main page, if you click on 'License', 'Support', etc. the title remains considerably hidden: 
Screenshot 2021-08-23 at 17 32 17

This was a problem earlier too, but less so:
Screenshot 2021-08-23 at 17 33 20

Formatting of code blocks on `/doc` is broken - compare a locally running version and sequenceserver.com/doc side by side

Indentation of stats-banner.html is inconsistent with the rest

discoform:
Thanks @yeban 

Yes, the bootstrap update came with some slight layout changes. I tried to keep everything as consistent as possible, but right now we're just trying to do the least effort migration to the latest version. I have a separate conversation going with Yannick about updating the markup of the sites (this and wumlab) which will be a larger effort, but will help a lot with cleaning up and maintaining the layouts.

I'll itemize the changes and make the adjustments to the titles and formatting of the code blocks and follow up.

PR #549, opened 2021-09-23
Set database title from text file

3lviend:
Update load extension, extend make_db_title method, fix get_format method to load only extracted fasta files

### DIFF
--- a/lib/sequenceserver/makeblastdb.rb
+++ b/lib/sequenceserver/makeblastdb.rb
@@ -138,7 +138,7 @@ module SequenceServer
     end
 
     def get_format(path, type)
-      exts = Dir["#{path}.*"].map { |p| p.split('.').last }.sort
+      exts = Dir["#{path}.#{type[0]}*"].map { |p| p.split('.').last }.sort
       EXPECTED_EXTENSIONS[type][exts] || []
     end
 
@@ -161,7 +161,7 @@ module SequenceServer
         next if @formatted_fastas.any? { |f| f[0] == path }
 
         @fastas_to_format << [path,
-          make_db_title(File.basename(path)),
+          make_db_title(path),
           guess_sequence_type_in_fasta(path)]
       end
     end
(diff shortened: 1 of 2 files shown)


lib/sequenceserver/makeblastdb.rb
```diff
@@ -138,8 +138,8 @@ def determine_formatted_fastas
     end
 
     def get_format(path, type)
-      exts = Dir["#{path}.*"].map { |p| p.split('.').last }.sort
-      EXPECTED_EXTENSIONS[type][exts] || []
+      exts = Dir["#{path}.#{type[0]}*"].map { |p| p.split('.').last }.sort
+      EXPECTED_EXTENSIONS[type][exts]
```
yeban, line 142: `|| []` in line 142 is important regardless of the change in line 141, so let's keep it.

lib/sequenceserver/makeblastdb.rb
```diff
@@ -161,7 +161,7 @@ def determine_unformatted_fastas
         next if @formatted_fastas.any? { |f| f[0] == path }
 
         @fastas_to_format << [path,
-          make_db_title(File.basename(path)),
+          make_db_title(path.clone),
```
yeban, line 164: Why do you need to clone path?

yeban (changes requested):

lib/sequenceserver/makeblastdb.rb
3lviend: @yeban I see, I'll revert it then

lib/sequenceserver/makeblastdb.rb
3lviend: @yeban because on super `make_db_title` method, i saw there is methods 'tr!', and 'gsub!', which is replace the current `path` variable value, and it will affect next call method `guess_sequence_type_in_fasta` and return an error. Should I clone it to different variable?

yeban: Hmm, what you need is “db_name = Fule.basename(path)” as first line of “make_db_title” method. Also, “db_name” argument of that method should be renamed to “path”.

3lviend: I see, I've updated the commit
Thanks

yeban:
Looking good. Do you mind signing the CLA (see first comment on the PR by CLAassistant)?

3lviend:
Yeah. Sure, I will

3lviend:
> Looking good. Do you mind signing the CLA (see first comment on the PR by CLAassistant)?

I did, but it still said not signed yet. not sure why.
or its require you to accept ?

Thanks

PR #579, opened 2022-03-21
Add "Share results" section to sidebar - Update sidebar.js

dpavam:
The sidebar has been updated with a new "share results" section and gives the user the option to copy the URL of their results into their clipboard.

A function was added that copies the URL into the user's clipboard (modified from: https://stackoverflow.com/a/49618964/18117380)

A new (sharing) panel was created and added to the JSX function. Within the panel, the copy URL functionality was included.
This was done to resemble the existing code as closely as possible.


yannickwurm:
Hi Diego - thanks for this. Note that you'll also want to use more informative commit & PR names. 

E.g., https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

yeban:
Hello,

I like this new feature. The implementation works.

I would request two compulsory change:
1. We use sentence case throughout, so 'Share Results' should be 'Share results'.
2. The newly introduced panel should have a different CSS class than 'downloads'. Maybe 'sharing-panel', or maybe you don't need a CSS class there. The reason to not use 'downloads' is that have some old jquery code that might get confused with two divs having the same class name (I am not sure it will as many parts were revised to be more robust but some doubt lingers).


Optionally, depending on the constraints you are working within, it would be nice to use another way to inform users that the url has been copied. Browser alerts are not a part of seqserv's design language. Many websites provide such feedback using a tooltip (e.g., copying repo url in GitHub) or other mechanisms. seqserv makes good use of bootstrap tooltips, so it would be a good fit here. Browser alerts introduce additional click to dismiss the notification.

It could be nice also to have a copy icon in front of the link
Screenshot 2022-03-28 at 17 37 32
.

yannickwurm:
Bumping issues with this one. 

Can you also provide relevant unit tests for this?

PR #582, opened 2022-04-12
Update job.rb

iQuxLE:
The job.yaml file has been updated with a separate entry for num_threads and query_length which includes the number of residues in query. Additionally, the total_database_length as ncharacters (inside the databases) has been added.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,4 +1,4 @@
 --color
 --backtrace
 --require spec_helper.rb
---pattern="spec/*_spec.rb","spec/blast_versions/blast_2.9.0/*.rb"
+
--- a/lib/sequenceserver/blast/job.rb
+++ b/lib/sequenceserver/blast/job.rb
@@ -115,7 +115,7 @@ module SequenceServer
         size = 0
         IO.foreach(@qfile) do |line|
           next if line[0] == '>'
-          size += line.strip.length
+          size += line.gsub(/\s+/, '').length
         end
         size
       end
--- a/lib/sequenceserver/blast/job.rb
+++ b/lib/sequenceserver/blast/job.rb
@@ -26,7 +26,7 @@ module SequenceServer
             @options   = @advanced + defaults
             @num_threads = Integer(config[:num_threads])
             @query_length = File.size(@qfile)
-            @total_database_length = database_ncharacter_length
+            @total_database_length = database_lengths
           end
         end
       end
@@ -60,7 +60,7 @@ module SequenceServer
                      " -query '#{qfile}' #{options}"
       end
 
-      def database_ncharacter_length
+      def database_lengths
         total_database_length_arr = databases.map(&:ncharacters)
         total_database_length_arr.map(&:to_i).reduce(:+)
       end
--- a/lib/sequenceserver/blast/job.rb
+++ b/lib/sequenceserver/blast/job.rb
@@ -24,9 +24,10 @@ module SequenceServer
             @databases = Database[params[:databases]]
             @advanced  = params[:advanced].to_s.strip
             @options   = @advanced + defaults
+            # The following params are for analytics only
             @num_threads = config[:num_threads]
-            @query_length = query_size
-            @total_database_length = ncharacters_total
+            @query_length = calculate_query_size
+            @databases_ncharacters_total = calculate_databases_ncharacters_total
           end
         end
       end
@@ -35,12 +36,12 @@ module SequenceServer
       # Attributes used by us - should be considered private.
       attr_reader :advanced
       attr_reader :databases
+      attr_reader :databases_ncharacters_total
       attr_reader :method
       attr_reader :num_threads
       attr_reader :options
-      attr_reader :query_length
       attr_reader :qfile
-      attr_reader :total_database_length
+      attr_reader :query_length
 
       # :nodoc:
       # Deprecated; see Report#extract_params
@@ -58,19 +59,6 @@ module SequenceServer
       def command
         @command ||= "#{method} -db '#{databases.map(&:name).join(' ')}'" \
                      " -query '#{qfile}' #{options}"
-      end
-
-      def ncharacters_total
-        databases.map(&:ncharacters).map(&:to_i).reduce(:+)
-      end
-
-      def query_size
-        size = 0
-        IO.foreach(@qfile) do |line|
-          next if line[0] == '>'
-          size += line.strip.length
-        end
-        size
       end
 
       # Override Job#raise! to raise specific API errors based on exitstatus
@@ -118,6 +106,19 @@ module SequenceServer
       # rubocop:enable Metrics/CyclomaticComplexity
 
       private
+
+      def calculate_databases_ncharacters_total
+        databases.map(&:ncharacters).map(&:to_i).reduce(:+)
+      end
+
+      def calculate_query_size
+        size = 0
+        IO.foreach(@qfile) do |line|
+          next if line[0] == '>'
+          size += line.strip.length
+        end
+        size
+      end
 
       def validate(params)
         validate_method params[:method]
--- a/lib/sequenceserver/blast/job.rb
+++ b/lib/sequenceserver/blast/job.rb
@@ -72,7 +72,6 @@ module SequenceServer
         end
         size
       end
-      
       # Override Job#raise! to raise specific API errors based on exitstatus
       # and using contents of stderr to provide context about the error.
       #
--- a/lib/sequenceserver/blast/job.rb
+++ b/lib/sequenceserver/blast/job.rb
@@ -72,7 +72,7 @@ module SequenceServer
         end
         size
       end
-      
+
       # Override Job#raise! to raise specific API errors based on exitstatus
       # and using contents of stderr to provide context about the error.
       #
@@ -142,7 +142,6 @@ module SequenceServer
       end
 
       def validate_databases(database_ids)
-        ids = Database.ids
         return true if database_ids.is_a?(Array) && !database_ids.empty? &&
                        (ids & database_ids).length == database_ids.length
         fail InputError, "Database id should be one of: #{ids.join("\n")}."
--- a/lib/sequenceserver/blast/job.rb
+++ b/lib/sequenceserver/blast/job.rb
@@ -24,7 +24,7 @@ module SequenceServer
             @databases = Database[params[:databases]]
             @advanced  = params[:advanced].to_s.strip
             @options   = @advanced + defaults
-            @num_threads = Integer(config[:num_threads])
+            @num_threads = config[:num_threads]
             @query_length = File.size(@qfile)
             @total_database_length = database_lengths
           end
--- a/lib/sequenceserver/blast/job.rb
+++ b/lib/sequenceserver/blast/job.rb
@@ -24,13 +24,16 @@ module SequenceServer
             @databases = Database[params[:databases]]
             @advanced  = params[:advanced].to_s.strip
             @options   = @advanced + defaults
+            @num_threads = Integer(config[:num_threads])
+            @query_length = File.size(@qfile)
+            @total_database_length = database_ncharacter_length
           end
         end
       end
 
       # :nodoc:
       # Attributes used by us - should be considered private.
-      attr_reader :method, :qfile, :databases, :advanced, :options
+      attr_reader :method, :qfile, :databases, :advanced, :options, :num_threads, :query_length, :total_database_length
 
       # :nodoc:
       # Deprecated; see Report#extract_params
@@ -48,6 +51,11 @@ module SequenceServer
       def command
         @command ||= "#{method} -db '#{databases.map(&:name).join(' ')}'" \
                      " -query '#{qfile}' #{options}"
+      end
+
+      def database_ncharacter_length
+        total_database_length_arr = databases.map(&:ncharacters)
+        total_database_length = total_database_length_arr.map(&:to_i).reduce(:+) 
       end
 
       # Override Job#raise! to raise specific API errors based on exitstatus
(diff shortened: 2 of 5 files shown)


lib/sequenceserver/blast/job.rb
```diff
@@ -50,6 +60,11 @@ def command
                      " -query '#{qfile}' #{options}"
       end
 
+      def database_ncharacter_length
+        total_database_length_arr = databases.map(&:ncharacters)
```
yannickwurm, line 64: would calling this database_lengths be more appropriate @iQuxLE ?

iQuxLE: Sure, I'll add that.

yannickwurm:
Is the choice of `integer()` once and `to_i` the other time intentional?

iQuxLE:
@yannickwurm I used `integer()` since that was used previously to call `config[:num_threads]` in sequenceserver.rb but I could cut it. It does not need to be converted.

`to_i` is actually needed in `database_ncharacter_length()` since `:ncharacters` is a string of numbers and I expect that to always be numbers and no other chars so I suggest thats fine.

yannickwurm:
if not needed then lets keep things simpler.

iQuxLE:
I updated that

yannickwurm:
Thanks @iQuxLE   reminder re my previous comment about naming. 

Also, what do you need to be able to add a/some unit test(s) to see that this is functioning correctly?

lib/sequenceserver/blast/job.rb
yeban: It is not clear if database_length refers to number of sequences in the database or number of characters.
- I think `ncharacters_total` might be a more appropriate name for your method.
- I don't think you need a variable within the method.


For context, we borrow as much of blast's terminology as possible. `ncharacters` is obtained from `blastdbcmd`'s output and it probably uses the same name.

yeban (changes requested):
The comment about refactoring is optional. The remaining two are required, in my understanding.

yannickwurm:
File.size is also different from actual query_length (for which we would also need to exclude identifier/descriptor lines).

I'm unsure whether the cpu/memory inefficient approach here would be problematic (ruby's equivalent or grep -v '>' | eliminate whitespace. length )

yeban:
> I'm unsure whether the cpu/memory inefficient approach here would be problematic (ruby's equivalent or grep -v '>' | eliminate whitespace. length )

Something like this would be memory efficient:

```
def query_size
  size = 0
  IO.foreach(file_name) do |line|
    next if line[0] == '>'
    size = size + line.strip
  end
  size
end
```

lib/sequenceserver/blast/job.rb
```diff
@@ -50,6 +60,19 @@ def command
...
+
+      def query_size
+        size = 0
+        IO.foreach(@qfile) do |line|
+          next if line[0] == '>'
+          size += line.strip.length
+        end
+        size
+      end
+
```
yeban, line 75: Can we make the two methods private and call them `calculate_databases_ncharacters_total` and `calculate_query_size`.

lib/sequenceserver/blast/job.rb
```diff
@@ -24,13 +24,23 @@ def initialize(params)
             @databases = Database[params[:databases]]
             @advanced  = params[:advanced].to_s.strip
             @options   = @advanced + defaults
+            @num_threads = config[:num_threads]
+            @query_length = query_size
+            @total_database_length = ncharacters_total
```
yeban, line 29: Can we add a comment above this code block indicating these are for analytics and not per se required to run a job.

yeban, line 29: As mentioned earlier, it is not clear if database_length means number of sequences or number of characters. A better name is `@databases_ncharacters_total`.

spec/job_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+            @params = {
+                sequence: 'AGCTAGCTAGCT', 
+                databases: [Database.first.id],
+                method: (Database.first.id == 'protein' ? 'blastp' : 'blastn'),
+            }
+        end
+
+        context 'with params' do
+            let(:test_job){test_job = Job.create(@params)}
```
yeban, line 49: Does this file pass rubocop check?

spec/job_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+                expect("#{__dir__}/dotdir/#{test_job.id}/job.yaml").not_to be_empty
+            end
+
+            it 'should create a stderr' do
+                expect("#{__dir__}/dotdir/*/stderr").not_to be_empty
+            end
+
+            it 'should create a stdout' do
+                expect("#{__dir__}/dotdir/*/stdout").not_to be_empty
+            end
```
yeban, line 62: Please remove these as they are not correct and it is not important for you to cover all aspects of job class as part of this pull-request.

spec/job_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+            it 'should have total database length' do
+                total_database_length = test_job.ncharacters_total
+                expect(total_database_length).to eql(total_database_length)
+            end
```
yeban, line 67: How do you justify this? You are saying a = b, thus we expect a = a.

spec/job_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+        # Test Job class from module Blast and params
+        ENV['RACK_ENV'] = 'test'
+        include Rack::Test::Methods
+    
+        before do
+            SequenceServer.init(database_dir: "#{__dir__}/database/v5/sample")
+
+            @params = {
+                sequence: 'AGCTAGCTAGCT',
```
yeban, line 42: Can we have a more realistic example sequence that captures new lines and sequence identifiers?

```
>seq1
ATGCATGCATGCATGC
>seq2
ATGCATGCATGCATGC
```

spec/job_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+                expect("#{__dir__}/dotdir/*/stdout").not_to be_empty
+            end
+
+            it 'should have total database length' do
+                total_database_length = test_job.ncharacters_total
+                expect(total_database_length).to eql(total_database_length)
+            end
+                
+            it 'should compute query length' do
+                test_job.query_size
```
yeban, line 70: Why is line 70 necessary?

spec/job_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+                expect(test_job.query_size).to eq(12)
+            end
+
+            it 'should be able to store a sequence in query.fa' do
+                test_job.send("store", 'query.fa', 'ACTA')
+                expect(test_job.query_size).to eq(4)
+                expect("#{__dir__}/dotdir/*/query.fa").not_to be_empty
+            end
```
yeban, line 78: Can we remove this please? It's neither correct, nor required.

spec/job_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+            it 'should have a command' do
+                expect(test_job.command).to start_with('blastn -db')
+            end
+            
+            it 'should have a method blastn when AGCTAGCTAGCT ' do
+                expect(@params[:method]).to eq('blastn')
+            end
```
yeban, line 86: Can we remove this please? It's neither correct, nor required.

yeban (changes requested):
Hello. You have made good progress on the PR. Good job for figuring out how to test things. I have suggested a few more changes that are required for me to okay the pull-request. Roughly, the comments are aimed at preventing multiple sources of truth in the API, better naming, trimming and making minor corrections to the test suite to keep it precise. I hope it is helpful.

spec/job_spec.rb
iQuxLE: Can you point out the mistakes I made? If it would be needed. How would I implement it?

spec/job_spec.rb
iQuxLE: With this I wanted to test that the private store method from Job is working correctly. But since this is not part of the PR I will delete it.

spec/job_spec.rb
iQuxLE: My mistake. It is not. I delete this.

spec/job_spec.rb
iQuxLE: Yes that's a mistake. Can I input the computed database length from job.yaml? But then I thought that I might have problems when the function is really not working correctly?
I could also just test for datatype (Integer). But that is also not very explicit.

Will the databases used for RSpec testing will always be the same? When these get updated I would run into the next problem if I just test for a specific value here, correct?

spec/job_spec.rb
iQuxLE: Could you explain why it is not correct?

spec/job_spec.rb
iQuxLE: I wanted to say yes. But I am not very sure since I was just expecting this when it passed all general checks here. Let me check this. I come back to this later.

spec/job_spec.rb
iQuxLE: Sure!

lib/sequenceserver/blast/job.rb
iQuxLE: One question. Isn't it bad practice to test private methods? I know that I did it in one part as well but I just got a little deeper. But if you think that is ok for this situation, I will get this implemented.

lib/sequenceserver/blast/job.rb
iQuxLE: Ok I forgot that, I'll update that.

iQuxLE: Sure, I think that's very helpful.

iQuxLE:
Thank you @yeban the comments are very helpful. At some points I would need maybe a bit more insights. I'll update everything asap.

lib/sequenceserver/blast/job.rb
yeban: But you don't need to test/call the private methods in your test cases. You have called them already in the constructor and saved the return value in instance variables that are accessible through instance attr_accessors.

spec/job_spec.rb
yeban: I think I mentioned it earlier. The test is not specific due to the wildcard (`*`). It can match stdout, stderr file of another job. Besides, I am not sure the files are created before the job is run.

spec/job_spec.rb
yeban: > Will the databases used for RSpec testing will always be the same? When these get updated I would run into the next problem if I just test for a specific value here, correct?

I think this is good question. Some databases will always stay part of the test suite, like the ones in `spec/database/v5/sample` except maybe uniprot/Swiss-prot subset. You can use them.

spec/job_spec.rb
yeban: `@params` is just a hash you created on line 41 of this file and set the method key to blastn based on database type. Is there a point of checking that? Besides, what does the test case description even mean?

iQuxLE: Ah ok right I see.

spec/job_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  describe 'BLAST Job' do
+    # Test Job class from module Blast and params
+    ENV['RACK_ENV'] = 'test'
+    include Rack::Test::Methods
+
+    before do
+      SequenceServer.init(database_dir: "#{__dir__}/database/v5/sample")
+
+      @params = {
+        sequence: ">seq1\nAGCTAGCTAGCT\n>seq2\nAATAGCTA",
```
yannickwurm, line 42: I suggest adding several more scenarios. 

E>.g, including spaces etc. (you're not testing the strip functionality)

.rspec
```diff
@@ -1,4 +1,4 @@
 --color
 --backtrace
 --require spec_helper.rb
---pattern="spec/*_spec.rb","spec/blast_versions/blast_2.9.0/*.rb"
```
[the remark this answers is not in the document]
iQuxLE: That can be in a separate PR, maybe?
It is needed for running test on specific files. That was not enabled before and very helpful for writing tests since otherwise the whole test-suite was always running during local testing with for example: `bundle exec rspec spec/job_spec.rb`

spec/job_spec.rb
iQuxLE: Ok that's doable.
What do you think of 
`@params = {
        sequence: ">seq1   \n AGCTAGCTAGCT \n>  SI2.2.0_80257 Si_gnF.scaffold" \
        "[2092364..2095011].pep_1  \n  AATAGCTA\n>s7:S11.0.9.1233 Fi_gnG.scaf" \
        "[5567] \n    PPKKLLMMOOPPPTTRYUIQWERTZ     ",`

Description, protein, and strip are included.

If whitespaces at the beginning of the fasta-header should be included, then the `calculate_query_length` function must be adjusted. The same goes for whitespaces in between the sequence. By now only leading and trailing whitespaces in a sequence are deleted.

I think we should leave '>' as the start of every fasta and additionally there shouldn't be a need of deleting whitespaces inside sequences as normal fasta files should not contain whitespace inside sequences. What do you think of that?

iQuxLE:
Oh no I apologise for this commit (Update sys.rb). This is not supposed to land here.

spec/job_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        databases: [Database.first.id],
+        method: (Database.first.id == 'protein' ? 'blastp' : 'blastn')
+      }
+    end
+
+    context 'with params' do
+      let(:test_job) { Job.create(@params) }
+
+      it 'should compute total characters of databases used' do
+        expect(test_job.databases_ncharacters_total).to eql(39_185_542)
```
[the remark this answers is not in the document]
iQuxLE: I added your suggestions. Please, check whether this is better now.

spec/job_spec.rb
iQuxLE: Rubocop checks are passing now, despite the very long describe block (should be under 25 lines). I checked with other files and we also have the problem there. For now I left it like that since it is not in the rubocop style guide and it is not disturbing regarding readability. If we would change that we should generally do that everywhere.

PR #583, opened 2022-04-19
Update React and convert react components to es6 class components syntax

### LINKED ISSUE #556: react is outdated
We should be using something much more recent

Lazy-poet:
fixes #556


yeban (changes requested):
Thank you. The PR is an incredible amount of effort and such modernisation of code is appreciated. It looks like the PR is work in progress? The results don't load when I run a BLAST search. Nor does capybara_spec.rb test suite pass.

yannickwurm:
Thank you @Lazy-poet - this is superb.

PR #590, opened 2022-05-15
Ruby 3.1.2 update

iQuxLE:
With these updates SequenceServer is able to work on Ruby 3.1.2


sequenceserver.gemspec
```diff
@@ -1,56 +1,58 @@
...
+s.license     = 'AGPL-3.0'
+
+s.summary     = 'BLAST search made easy!'
+s.description = <<~DESC
+  SequenceServer lets you rapidly set up a BLAST+ server with an intuitive
+  user interface for use locally or over the web.
+DESC
+
+# dependencies
+s.required_ruby_version = '>= 3.1.2'
```
yeban, line 19: I think we need to distinguish between ensuring seqserv works on latest Ruby and the minimum Ruby version we support. As a rule of thumb, the minimum Ruby version we want to support is the default Ruby of the oldest Ubuntu LTS that has not entered ESM phase yet. That would be Ubuntu 18.04 and the default for Ubuntu 18.04 is Ruby 2.5. So we should ideally have `>= 2.5.0` here.

sequenceserver.gemspec
```diff
@@ -1,56 +1,58 @@
...
+s.authors     = ['Anurag Priyam', 'Ben J Woodcroft', 'Vivek Rai', 'Yannick Wurm']
+s.email       = ['anurag08priyam@gmail.com', 'b.woodcroft@uq.edu.au']
+s.homepage    = 'http://sequenceserver.com'
+s.license     = 'AGPL-3.0'
+
+s.summary     = 'BLAST search made easy!'
+s.description = <<~DESC
+  SequenceServer lets you rapidly set up a BLAST+ server with an intuitive
+  user interface for use locally or over the web.
+DESC
```
yeban, line 16: Please undo whitespace changes throughout this file. Currently it looks like the whole file has been rewritten.

yeban (changes requested):
The first point about removing whitespace changes is essential. The second point probably requires some discussion.

sequenceserver.gemspec
yeban: Is it possible for us to support all the way from Ruby 2.5 to 3.1?

yannickwurm:
To keep things simple, I think we can decide that the next release is for 3.1. 

Yannick

sequenceserver.gemspec
yeban: I appreciate the reasoning that soon enough we will run into the next problem, but we should think from the user's point of view as well: Enforcing such a high minimum Ruby version constraint will make it harder for many existing users to upgrade and for new user to even try it out.

One way of approaching this is to test against multiple Ruby versions on Travis. Maybe just Ruby 2.5 and 3.1. That is, the minimum required and the latest. If installation and tests start failing for the minimum Ruby version and without a simple fix, we increase the minimum Ruby. It used to be the case earlier that we used to test against multiple Ruby versions. But there was a phase when our testing system went through a lot of changes. And it got too complex to test multiple Ruby versions while also experimenting with different approaches, so the number of Ruby versions to run the tests on was reduced. I think the test suite is in a much better position today (except 1-2 remaining random errors which I am sure we will get to as well).

iQuxLE:
The recent updates will make the sequenceserver usable for everyone from 2.5 to 3.1.2.

yannickwurm:
How can we get travis to run on this file?

yannickwurm:
(or perhaps we should just migrate to Github's built-in test runner - given that travis is also funny with timeouts)

yannickwurm:
@iQuxLE  - it looks like you haven't yet addressed some of the issues that have been raised

iQuxLE:
Gonna take care of this the next days.

yannickwurm:
Superseded by #637

PR #597, opened 2022-05-22
Minor updates to workflow due to the switch to webpack

yeban:
- Add a `watch` script to npm that runs `webpack --watch`.
- Accordingly update development instructions in README
- Update Dockerfile and .dockerignore so that docker build --target=minify continue to work (FYI @nathanweeks)
- Changes to package-lock.json after first run of npm install on my Mac after the switch to webpack


yannickwurm:
Super thank you

PR #591, opened 2022-05-20
Add "Share results" section to sidebar and "send URL by email" functionality. - Update sidebar.js

dpavam:
Adds a new "Share results section to the sidebar. Within this section, a "Send by email" button enables users to send the URL, search id, and the first 15 databases of their results through email. 

This also resolves some issues pointed out here: #579. A different CSS class ('sharing-panel) has been given to the sidebar and the sentence case has been preserved. The "Copy URL to clipboard" has been removed from this version. 

Unit tests check the contents of the "mailto" message.

Issue #568, opened 2021-12-19
Add "share" link on the left under downloads


yannickwurm:
It isn't obvious to users that a search result URL is stable (until deleted). 

To make this more obvious, we should add an explicit 
**Share results** section, with a "Copy URL to clipboard" text that does what it says

PR #600, opened 2022-05-25
Add "Share results" section to sidebar in ES6 version. Update sidebar.js

dpavam:
Includes features from #591 and #599 into the recent ES6 update. 

These cover: 

- Adds a new "Share results section to the sidebar. Within this section, a "Send by email" button enables users to send the URL, search id, and the first 15 databases of their results through email.
- Fixes tooltips in sidebar.js. Previously, only standard boxes would display messages instead of bootstrap tooltips.
- Addresses issue #568.
- Adds a "Copy URL to clipboard" functionality on the sidebar. Displays a tooltip as a confirmation to the user.

- [x] Note: the methods to copy are deprecated and using [clipboard API](https://developer.mozilla.org/es/docs/Web/API/Clipboard_API) is recommended. However, this is only supported under HTTPS, we currently use HTTP in the results section.
- [x] @Lazy-poet, I spoke with @yannickwurm. It might be a good idea to have your feedback on this new feature.
- [x] @yeban will have a look at the tooltip since it fails to reappear at times.


public/js/sidebar.js
```diff
@@ -16,10 +16,14 @@ export default class extends Component {
         this.downloadFastaOfSelected = this.downloadFastaOfSelected.bind(this);
         this.downloadAlignmentOfAll = this.downloadAlignmentOfAll.bind(this);
         this.downloadAlignmentOfSelected = this.downloadAlignmentOfSelected.bind(this);
+        this.componentDidMount = this.componentDidMount.bind(this);
```
Lazy-poet, line 19: componentDidMount is an inbuilt react method that doesn't need to be bound (especially since you aren't particularly calling it yourself)

dpavam: Removed it from the constructor

yeban:
@dpavam thanks, looking into it

yeban:
@dpavam 

The fact that the tooltip doesn't disappear automatically when you set `trigger: 'click'` seems to be Bootstrap 3's  default behaviour. I tested this by bypassing the wrapper `tooltip` method we have in `jquery_world.js` and got the same behaviour. So I think your approach of hiding the tooltip using `setTimeout` is fair.

Next, the tooltip doesn't show again after you have called `hide` on it once because the `tooltip` method is a wrapper around Bootstrap's `tooltip` that is defined in `jquery_world.js`. It doesn't behave exactly as Bootstrap's `tooltip` method. It destroys the previous tooltip instance and tries to create a new one. But I think it fails to create a new instance correctly because it expects an `options` object but receives the string `'hide'`.

Bootstrap 3's default `tooltip` is renamed to `_tooltip` in `jquery_world.js`. You can use it hide tooltip in your `setTimeout` instead:

```
-            $('#copyTooltip').tooltip('hide');},3000);
+            $('#copyTooltip')._tooltip('hide');},3000);
```

Furthermore, I suggest adding `delay: 0` when you initialise the tooltip instance in `componentDidMount`:

```
-            $('#copyTooltip').tooltip({title:'Copied!',trigger:'click', placement: 'right'});
+            $('#copyTooltip').tooltip({title:'Copied!',trigger:'click', placement: 'right', delay: 0});
```

It's not perfect still - the tooltip still won't show on alternate clicks. I am not quite sure why this happens, but I am inclined to think it is okay. It might be happening because the call to `select` in `copyURL` method changes focus to a hidden element and that confuses bootstrap.

As for your idea about showing 'Click to copy' on hover and change the text to 'Copied' on click, I tried but didn't manage to implement. I think it is not required either because the text of the link is self-explanatory (for example, we don't have tooltips for all links in the sidebar).

Tooltips aside, I think you should clean up your code a bit, for example, by running it through eslint. I will try to leave some more comments later in the evening.

dpavam:
Hi @yeban, thanks for the feedback. I have implemented the changes.
I ran eslint using codeclimate as:
 `codeclimate analyze -e eslint public/js/sidebar.js ` 

and the only issues were: 
`== public/js/sidebar.js (2 issues) ==`

`1: 'React' is defined but never used. [eslint]`
`90: Unexpected console statement. [eslint]`

Should I use eslint in some other way or is there a specific section that needs tidying up?

public/js/sidebar.js
```diff
@@ -100,13 +103,55 @@ export default class extends Component {
         return false;
     }
 
+    /**
+     * Fixes tooltips in the sidebar, allows tooltip display on click
+     */
+    componentDidMount() {
+        $(function () {
+            $('.sidebar [data-toggle="tooltip"]').tooltip({placement: 'right'});
+            $('#copyTooltip').tooltip({title:'Copied!',trigger:'click', placement: 'right',delay: 0});
```
yeban, line 112: Missing space after `:` and `,`.

public/js/sidebar.js
```diff
@@ -100,13 +103,55 @@ export default class extends Component {
...
+        });
+    }
+
+    /**
+     * Handles copying the URL into the user's clipboard. Modified from: https://stackoverflow.com/a/49618964/18117380
+     * Hides the 'Copied!' tooltip after 3 seconds 
+     */
+
+    copyURL() {
+        var element = document.createElement('input'), url = window.location.href;
```
yeban, line 122: I find multiple variable assignment on a single line quite confusing. In the minimum, break line after `,`. Better, use two `var` statements.

public/js/sidebar.js
```diff
@@ -100,13 +103,55 @@ export default class extends Component {
...
+        document.body.appendChild(element);
+        element.value = url;
+        element.select();
+        document.execCommand('copy');
+        document.body.removeChild(element);
+        
+        setTimeout(function(){
+            $('#copyTooltip')._tooltip('hide');},3000);
```
yeban, line 130:
```js
setTimeout(function() {
  $('#copyTooltip')._tooltip('hide');
}, 3000);
```

or

```js
setTimeout(() => $('#copyTooltip')._tooltip('hide'), 3000)
```

Note the differences with your version.

public/js/sidebar.js
```diff
@@ -231,11 +276,47 @@ export default class extends Component {
...
+                        <li>
+                            <a id="sendEmail" className ="btn-link email-URL" data-toggle="tooltip"
+                                title="Send by email" href= {`mailto:?subject=SeqServ results &body=Thank you for using SequenceServer.
+                                %0DBelow, you will find a link to the results of your most recent search. While using SequenceServer, you may use this link to access previous results.
+                                %0DYou will also find the unique ID of your query and the first 15 databases used in your search.
+                                %0D%0DLink: `+ window.location.href + '%0DQuery id: ' + this.props.data.search_id +
+                                '%0DDatabases:'  + this.usedDatabases() + '%0D%0DPlease cite: https://doi.org/10.1093/molbev/msz185'}
+                                target="_blank" rel="noopener noreferrer">
+                                <i className="fa fa-envelope"></i> Send by email
+                            </a>
```
yeban, line 307: I think you are better off extracting href's value into a function of its own (maybe call it `mailtoLink`. Use multiline string in the function and `urlQuoteParams` or other suitable function to automatically turn special characters to `%0D` etc.

yeban (changes requested):
I thought eslint was supposed to catch some of the syntax issues I highlighted, but maybe not. I use Prettier extension for VS Code to prettify code when I am uncertain. There is a shortcut (which I can't remember because I have Vim keybindings) that you can press and the extension reformats the line you are on. You can also select a block of code and hit this extension - the whole block will be formatted.

public/css/sequenceserver.css
```diff
@@ -583,6 +583,10 @@ input[type=checkbox] {
   cursor: pointer;
 }
 
+.sharing-panel a {
+  cursor: pointer;
+}
+
```
yeban, line 589: I grepped for `cursor: pointer` in our `public/css` folder and came up with the following:

```
.cursor-pointer {
  cursor: pointer;
}
```

Thus an alternative is that you add the `cursor-pointer` CSS class to your `a` elements.

public/js/sidebar.js
dpavam: Initially, ` href ` was in a function that was assigned to onClick in this `a` element. It was changed to the current format to access the `href` for testing (Capybara_rspec.rb, 222). I can revert as suggested but could you advise on how to check the `href` contents in the function for testing?

public/css/sequenceserver.css
dpavam: For clarification, would that be something like: ` className="btn-link email-URL cursor-pointer"`?

public/js/sidebar.js
yeban: Not sure I fully understand your concern. You would access href in tests like you do now. You don’t need to assign href on click. It’s a fixed string for a given results page. You just construct it in a function and assign the return value to href.

    mailtoLink () {
      return ‘mailto:blah’
    }

    <a href={this.mailtoLink()}>Share</a>

public/css/sequenceserver.css
yeban: Yes

public/js/sidebar.js
yeban: Did you inspect the output of `encodeURI` to see if there is `%09` in it?

Maybe this is what you wanted: `message.replace(/(%20){2,}/g, '')`.

But maybe npm packages like dedent are also a good alternative.

yannickwurm:
Hi @dpavam  - could you please check what the spec conflict is? I Guess we need to manually rebuild the min.js too
Thanks!

PR #601, opened 2022-05-26
Clear SequenceServer::Database.collection after each test context

yeban:
Test suites may initialise SequenceServer using different databases
directories in spec/. Many of the databases in the different subdirs are
essentially the same (same content, same display name), just formatted a
little different: v4, v5, with or without -parse_seqids option. This can
result in an accumulation of databases with the same name in the process
and cause confusion. For example, Capybara test that require selecting a
database in the interface based on name may fail because of ambiguous
matches.

Automatically clearing SequenceServer::Database.collection fixes this
problem.

/cc @iQuxLE

### DIFF
--- a/spec/download_helper.rb
+++ b/spec/download_helper.rb
@@ -14,7 +14,8 @@ module DownloadHelpers
   end
 
   def downloaded_file
-    downloads.first
+    warn "*** Multiple files in downloads directory. Expected one. ***" if downloads.length > 1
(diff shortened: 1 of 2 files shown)


yannickwurm:
Screenshot 2022-05-29 at 15 52 35

yeban:
It is one of those random failures on Travis, but it might be possible to prevent it (f99e662).

yannickwurm:
Awesome thank you

PR #603, opened 2022-06-05
Hsp alignment download reformat

iQuxLE:
FASTA headers of the downloaded HSP alignments now show the hit number and hsp number (in letters) that the user is currently looking at.


yannickwurm:
Hey @iQuxLE - do these code climate need to be fixed? or are they minor?

And/or are they related to the conflicts that need to be resolved. Are you able to resolve those?

iQuxLE:
I suppose the issues are more minor. It is just a duplication in two locations and minimizing that won't make it more intuitive and easier to understand.

yannickwurm:
Are you able to tell Codeclimate to ignore those minor issues (through local in-line ignore/reconfig comments?)

iQuxLE:
I added filters for the specific functions inside alignment_exporter.js and sidebar.js
These filters use the parsing tree to allow codeclimate to prevent a duplication call as these are minor duplications.
By that we don't need to change the mass_threshold which would affect all files in sequenceserver.

yeban:
Given the number of 'Update xyz' kind of commits, it may be desirable to squash-and-merge this PR. That would be `git merge --squash` on the command line, but also GitHub would offer this option when the remaining file conflict is resolved.

yannickwurm:
@iQuxLE  
Screenshot 2022-07-31 at 22 31 31

iQuxLE:
I think I have not used the most recent commit that's why there is a conflict cause I don't see the details for this. Let me check the next days.

yannickwurm:
Yes I think we need a Github action that automatically rebuilds minimised js and css whenever the repo moves

yannickwurm:
bump

yannickwurm:
@augeas  - could you please pull out the essentials from here?

PR #609, opened 2022-07-12
Sticky notification for mixed sequences

Lazy-poet:


yannickwurm:
Hi @Lazy-poet  - when I try this, the mixed notification still disappears. 

Some of this is due to clicking elsewhere in the window. We want this 'mixed' notification to stay even when user has clicked elsewhere.e

PR #611, opened 2022-07-12
add previous and next buttons to navigate longer queries

### LINKED ISSUE #592: When more than 8? queries, should show a "next query" button
When many queries, we don't show an index (for obvious reasons). But it would still be handy to be able to navigate to next.
"next" and "previous" buttons

Screenshot 2022-05-20 at 21 13 18

Lazy-poet:
This PR fixes #592


yannickwurm:
Thanks @Lazy-poet  - can we have tests for this as well?

yannickwurm:
Three additional things:
- [ ]  it doesn't "know" where it is. If I scroll down to the 10th one, then hit next, it takes me to the 2nd rather than the 11th. 
- [ ] The "previous" should not be shown if we are at the top of the page. The "next" should not be shown if we've passed the last one. (i.e. not shown - and the | not shown either - rather than simply disabling). 
- [ ] I think we want a bit more vertical space here too. It looks imbalanced compared to what's happening at Download section of the sidebar.  Screenshot 2022-07-31 at 22 50 47

Lazy-poet:
> Three additional things:
> 
> * [ ]   it doesn't "know" where it is. If I scroll down to the 10th one, then hit next, it takes me to the 2nd rather than the 11th.
> * [ ]  The "previous" should not be shown if we are at the top of the page. The "next" should not be shown if we've passed the last one. (i.e. not shown - and the | not shown either - rather than simply disabling).
> * [ ]  I think we want a bit more vertical space here too. It looks imbalanced compared to what's happening at Download section of the sidebar.  Screenshot 2022-07-31 at 22 50 47
> * [ ]  Automated test of the transition - when do we have the full index, and when do we show previous/next buttons.
> * [ ]  Automated test of the scrolling behavior (it should know where it is

@yannickwurm My most recent commit fixes 1-3 of this comment. Tests for these behaviors are underway and will be updated soonest.

Lazy-poet:
the last commit address #647

PR #612, opened 2022-07-12
prevent text wrapping in hsp headers

### LINKED ISSUE #593: inappropriate text wrapping in header of HSPs
Text ends up being too long and wrapping. 
(in example db (?) this is from tblastx of SiJWA01ABG.scf 

Instead it would be preferable to change the kerning or distance so the long line fits into one line. Or remove some spaces. 
Using a non-monospace font may work for this.  (we only need the monospace for the actual alignment). [e.g., the "Input" font has mono-space and non-monospaced variants)

Screenshot 2022-05-20 at 21 16 52

Lazy-poet:
This PR fixes #593


yannickwurm:
Thank you @Lazy-poet . Could you clarify why we need additional packages such as `file-loader`? Thank you very much

Lazy-poet:
> Thank you @Lazy-poet . Could you clarify why we need additional packages such as `file-loader`? Thank you very much

It was one of the plugins I used in webpack while trying to use webpack to minify the css, but I realized there was a better alternative, clean-css and I forgot to remove it. will remove it now and push changes

PR #618, opened 2022-08-10
Add "Share results" section to sidebar

dpavam:
Follows pull request #600 and resolves conflicts

PR #620, opened 2022-08-12
Removes unnecessary code in copyURL 

dpavam:
Removes an unnecessary line introduced in #618.


yannickwurm:
Hi @dpavam  - this PR looks like it does much more than remove an excess line. 

Can you split this into 2 please?

Thanks,
Yannick

dpavam:
That is right, I did not notice the changes from the other featured leaked into this one. I will change it.

PR #629, opened 2022-11-19
Issue/615

Lazy-poet:


augeas:
`README.md` neglects to instruct to install Jest: `pm install --save-dev jest`

There *are* UI tests, and they *do* pass.

There's is "debate" in the React community WRT Jest vs. Mocha, Mocha is more popular, but Jest is more official. It seems unlikely that anyone else in the code-base will have a strong enough opinion to reject the use of Jest.

Seems good to merge.

PR #656, opened 2023-06-19
Fix alignments download bug on safari

Lazy-poet:
Safari doesn't support programmatically clicking a link to prompt download unless the click is initiated by an actual user. Previous code implementation for downloading alignments waits for the user to first click the `Alignment of selected hits` link before it starts generating the alignments, then it creates a temporary URL for downloading the alignment file and then creates an offscreen anchor element whose href is set to the URL created which is then clicked programmatically to begin download. This process works normally on all browsers asides Safari (because the final click didn't come from the user), so I implemented a workaround whereby the alignments are generated as soon as the user checks the 'select' boxes and then the url for downloaded is directly appended to the `Alignment of selected hits` link so that when the user clicks on this link, downloads starts immediately.

Similarly, for the `Alignment of all hits` link, the download URL is generated and appended to the respective link element as soon as the component mounts and queries are fetched.


yannickwurm:
Note: all of this is client-side (no server-calls)

PR #676, opened 2023-08-04
Test cases grg

augeas:
This updates the "tests" workflow to merge the coverage report from the new js tests with the rspec coverage and sends both to CodeClimate, also attaching both to the action annotations.

The Jest configuration, as supplied, was including several spurious files in the coverage reports, e.g:

`Users/priyam/sequenceserver/node_modules/biojs-vis-sequence/node_modules/biojs-events`

(See [previous executions](https://github.com/wurmlab/sequenceserver/actions/runs/5693685820/job/15433297784#step:12:27) of the action. )

Although the value of `collectCoverageFrom` in `jest.config.js` looks entirely reasonable, removing it and reverting to defaults seems to have resolved the issue.

The additions to the workflow can be seen running successfully on the [`actions-grg`](https://github.com/wurmlab/sequenceserver/actions/runs/5739775615/job/15556154343) branch.


augeas:
Killed off some of the `page.executes` that were using jQuery in the rspec tests, this reduces but doesn't entirely eliminate the random test failures.

Issue #669, opened 2023-07-20
Number of queries stats in job.yaml


tadast:
Currently, Sequenceserver stores the [total query length in job.yaml as query_length](https://github.com/wurmlab/sequenceserver/blob/37b3cb13da870b770a2bdb6e0d4a0a7c5b9fee67/lib/sequenceserver/blast/job.rb#L30).

Add another datapoint that records the number of query sequences. This can be obtained by looking into the query (which ultimately is saved as `query.fa`, but hopefully this datapoint can be captured in-transit?). 

The value would be a number of lines starting with `>` in the query in an ideal scenario. However, edge cases exist, when the `>` is omitted, e.g.:
- if there are no lines starting with a `>`, the whole query is a single sequence
- if the first line does not start with a `>`, the number of queries would be `number of lines starting with > + 1` e.g. [this query](https://github.com/wurmlab/sequenceserver/blob/bf24a05614e0b92d69b22e98914288214dd99c32/spec/sequences/protein_query.fa) contains 2 sequences

PR #679, opened 2023-08-11
add method for computing number of sequences in a query

Lazy-poet:
This addresses issue #669

### DIFF
--- a/lib/sequenceserver/blast/job.rb
+++ b/lib/sequenceserver/blast/job.rb
@@ -21,6 +21,7 @@ module SequenceServer
           validate params
           super do
             @method    = params[:method]
+            @query = params[:sequence]
             @qfile     = store('query.fa', params[:sequence])
             @databases = Database[params[:databases]]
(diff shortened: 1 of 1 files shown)


lib/sequenceserver/blast/job.rb
```diff
@@ -20,23 +20,32 @@ def initialize(params)
...
+            @method = params[:method]
+            @query = params[:sequence]
             @qfile     = store('query.fa', params[:sequence])
             @databases = Database[params[:databases]]
             @advanced  = params[:advanced].to_s.strip
             @options   = @advanced + defaults
             # The following params are for analytics only
             @num_threads = config[:num_threads]
             @query_length = calculate_query_size
+            @number_of_sequences = calculate_number_of_sequences
```
tadast, line 32: I think it would be more self explanatory if it was called number_of_query_sequences or sequences_in_query, what do you think?

tadast (commented):
Variable name could be improved, but the code looks good to me :+1:

Issue #666, opened 2023-07-19
Job 404 should not throw an exception


tadast:
When a job is not found i.e. when visiting `<sequenceserver-hostname>/this-job-definitelly-does-not-exist` an exception is thrown. This results in exception notifications for those hosting sequenceserver instances. 404 not found is not an exceptional case, so perhaps it should be handled at the action level with a `halt 404` and optionally rendering a dedicated 404 page.

<img width="1068" alt="image" src="https://github.com/wurmlab/sequenceserver/assets/218239/06d80d38-dd29-40ab-ac0d-4a0be8568df0">

PR #710, opened 2023-12-07
Fixed the job 404 exception error #666

gorsyan:
In` job.rb` file `return nil` if the job doesn't exist.
In `routes.rb` file where the `Job.fetch(id)` method is called, handle the scenario when `Job.fetch` returns `nil` and add corresponding `halt 404`.


yannickwurm:
Thank you @gorsyan

Issue #712, opened 2023-12-20
Sidebar Element Click Causes Exception and User Work Loss


gorsyan:
### Issue Overview
When attempting to interact with a sidebar element that is supposed to be non-clickable, it unexpectedly responds to the click. This interaction leads to an exception and subsequently redirects to an error page. This abrupt redirection results in the loss of any unsaved work by the user.

yannickwurm:
Thanks @gorsyan  - can you be more explicit? Make a loom recording?

yannickwurm:
Ok - I've reproduced - thank you for catching this. To reproduce: 
1. hover over FASTA of selected hits.  The pointer changes to a "cannot click" symbol
2. Click anyways

Seqserv receives this click. And errors with "CommandFailed" - likely because the get fasta function was called without the required params (identifiers of sequences).

PR #714, opened 2023-12-24
Resolved Sidebar Clickability Issue for Non-Interactive Elements

gorsyan:
Fixed Bug #712  Corrected an Issue Where the Sidebar Element Remained Clickable Despite Being Intended as Non-Interactive, Leading to User Data Loss. Implemented Explicit CSS Restrictions to Disable Pointer Events on Elements with the 'Disabled' Attribute. Additionally, Added a Check to Ensure Event Listeners in the Sidebar Element are Inactive if the Array of Selected Items is Zero, Preventing Unintended Interactions.


public/css/sequenceserver.css
```diff
@@ -84,6 +84,8 @@ h4 > .fa {
  */
 .disabled {
   color: #ccc !important;
+  cursor: not-allowed;
+   pointer-events: none; /* This prevents the click from happening */
```
yannickwurm, line 88: Spacing seems to be off on line 88

yannickwurm:
Hiya @gorsyan  - I'm sorry I suspect that this may now need to be rebased given that there have been minor changes to what contributes to the .min.js

PR #733, opened 2024-03-13
Feat: histogram graph

joko3ono:
1. add null plugins for histogram graph
2. add alias to grapher so we can call it from extensions


tadast (approved):
LGTM, codeclimate issue can be ignored, it's getting confused with jsx in js files

PR #743, opened 2024-05-27
Update Dockerfile for BLAST 2.15.0

nathanweeks:
SequenceServer 3.0.1 supports BLAST 2.15.0, though the Dockerfile currently specifies 2.14.0:

https://github.com/wurmlab/sequenceserver/blob/7233794023dc41ee1c5379705cb67504d7d47479/Dockerfile#L3

This PR proposes updating the Dockerfile for BLAST 2.15.0, with several other ancillary modifications:

* Facilitate arm64 container builds by using [ncbi/blast-static](https://hub.docker.com/r/ncbi/blast-static/tags) (which has linux/arm64 images)
* Use ruby "bookworm" base images instead of "bullseye" to fix resulting `GLIBCXX_<version> not found` errors
* While devops best practices generally involve pinning dependencies as precisely as possible for reproducibility, suggest not pinning ruby patch version (i.e., (`ruby:3.2` instead of `ruby:3.2.<patch>`) so newer/patched versions are more easily/likely used during development & production builds 
* Update node base image from (EOL) version 15 to (LTS) version 20.
* Explicitly specify container registry for build environments where docker.io is not the default

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 # Build variables. These need to be declared befored the first FROM
 # for the variables to be accessible in FROM instruction.
-ARG BLAST_VERSION=2.14.0
+ARG BLAST_VERSION=2.15.0
 
 ## Stage 1: gem dependencies.
-FROM ruby:3.2.2-bullseye AS builder
+FROM docker.io/library/ruby:3.2-bookworm AS builder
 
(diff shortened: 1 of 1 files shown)


tadast (approved):
Looks great, thanks @nathanweeks 🙌

PR #747, opened 2024-06-11
Fix: flickering when load report

joko3ono:
### Context

When I created the histogram, I found that the diagram (circos and histogram) flickered for up to 4 minutes or more, when it flickers, I can't inspect the diagram because it keeps rendering.

https://github.com/wurmlab/sequenceserver/assets/1142773/3ef9af05-2caa-46bc-9c3a-8015780edea9

### Solution

In this PR, I separated the diagram and results rendering so that the diagram will only render once.

https://github.com/wurmlab/sequenceserver/assets/1142773/223efeb8-e054-445b-a902-3751df00e5a5


public/js/hits.js
```diff
@@ -0,0 +1,276 @@
...
+    * start iteratively adding 1 HSP to the page every 25 milli-seconds.
+    */
+    componentDidUpdate(prevProps, prevState) {
+        // Log to console how long the last update take?
+        // console.log((Date.now() - this.lastTimeStamp) / 1000);
+
+        // Lock sidebar in its position on the first update.
+        if (this.nextQuery == 0 && this.nextHit == 0 && this.nextHSP == 0) {
+            this.affixSidebar();
+        }
```
tadast, line 40: Should this be a responsibility of report.js? I think now with the custom hits batch-rendering logic separated, we should be able to affix the sidebar one in report.js without any of the batch-rendering related checks or am I missing something?

public/js/null_plugins/report_plugins.js
```diff
@@ -16,7 +17,12 @@ class ReportPlugins {
     }
 
     generateStats() {
-        return (<Histogram />);
+        var colors = chroma.scale('RdYlBu').colors(1);
+        return (
+          <div className='histogram-container'>
+            <Histogram query="evalue" name="E-Value Distribution" color={colors[0]} />
+          </div>
```
tadast, line 24: Would it be possible to clean up this PR to only inlcude the performance fix without the Histogram changes?

tadast (commented):
First of all, I really like that hits got their own component now, makes the code a bit easier to follow, thanks 🙌  Overall it looks good, but to get this over the finish line we need to fix the legitimate spec failures in the suite:

image

Also a couple suggestions in the comments.

public/js/hits.js
joko3ono: @tadast from my perspective it's better this way

in report js, we already have 2 checks
- if a warning present
- if results available

public/js/null_plugins/report_plugins.js
joko3ono: can we keep it?

this PR is required to fix performance before we merge histogram.

PR #765, opened 2024-07-09
[WIP] Update JS deps

tadast:
Primary goal: bump jQuery to v3 and Bootstrap without introducing any regressions.

Ideal scenario: Whatever is breaking in jQuery can be replaced by vanillaJS/React. Would be great to get rid of Bootstrap js dependencies too, if the effort makes sense.


tadast:
Closing in favour of #767

PR #767, opened 2024-07-13
Update JS deps

joko3ono:
What does this PR contain?

- Fix: jest error
- Fix: After search page, the loading page halted 
Screenshot 2024-07-15 at 01 56 19

- Fix: Misterious tests error
`SyntaxError: Unexpected token < in JSON at position 0 at parse (<anonymous>)`
- Resolve react warning: Support for string refs will be removed in a future major release. We recommend using useRef() or createRef() instead. Learn more about using refs safely here:
https://reactjs.org/link/strict-mode-string-ref
- Complete: React warning about ref
- Fix: Sequence View is not functioning
Screenshot 2024-07-15 at 06 22 05


public/js/sequence_modal.js
```diff
@@ -160,6 +150,11 @@ class SequenceViewer extends React.Component {
         footer: false,
       },
     });
-    widget.hideFormatSelector();
+    setTimeout(function() {
+      if (widget._headerDiv != null) {
+        widget.hideFormatSelector();
+        return;
+      }
+    }, 200);
```
tadast, line 158: What's happening here?

public/js/tests/report.spec.js
```diff
@@ -105,32 +123,36 @@ describe('REPORT PAGE', () => {
...
-                    const expected_num_hits = container.querySelectorAll('.hit-links input[type="checkbox"]').length;
-                    const file_name = `alignment-${expected_num_hits}_hits.txt`;
-                    expect(alignment_download_link.download).toEqual(file_name);
-                    expect(alignment_download_link.hred).not.toEqual('#');
+                    // const alignmentDownloadLink = container.querySelector('.download-alignment-of-all');
+                    // const hitLinks = container.querySelectorAll('.hit-links input[type="checkbox"]');
+                    // const expectedNumHits = hitLinks.length;
+                    // const fileName = `alignment-${expectedNumHits}_hits.txt`;
+                    // expect(alignmentDownloadLink.download).toEqual(fileName);
+                    // expect(alignmentDownloadLink.href).not.toEqual('#');
```
tadast, line 139: Is there a problem with these tests? Looks like a useful test

public/js/tests/cloud_share_modal.spec.js
```diff
@@ -0,0 +1,75 @@
...
+
+        // Wait for the loading state to finish
+        await waitFor(() => getByText('Uploading the job to SequenceServer Cloud, please wait...'));
+
+        // Check that the results are displayed
+        await waitFor(() => getByText('Network response was not ok'));
+
+        expect(getByText(/Network response was not ok/i)).toBeTruthy();
+    });
+});
```
tadast, line 75: Nce :+1:

public/js/sequence_modal.js
joko3ono: There is a race condition occurring when we invoke hideFormatSelector before _headerDiv is set.

public/js/tests/report.spec.js
joko3ono: oh it's my fault, I was debuging before each and forget to uncomment this one, will fix on the next PR

PR #776, opened 2024-08-02
Convert bootstrap to tailwind

3lviend:
- [x] Converted bootstrap to tailwind
- [x] Cleaning up unused Bootstrap CSS/JS code

Here is the recording:
https://www.loom.com/share/cb6c39b86ea14988ab2c14d7db13ed8a?sid=cdbb2359-06fa-49a1-976d-c42097b7ddaf


public/css/sequenceserver.css
```diff
@@ -507,3 +421,8 @@ td.nowrap-ellipsis {
   display: flex;
   justify-content: space-between;
 }
+
+/* override tailwind class  */
+.svg-container.collapse {
```
tadast, line 426: Would it be better to change this CSS class name in JSX instead of fighting tailwind?

tadast (commented):
Good progress :+1: 

If I recall correctly, the code was also relying on bootstrap for showing tooltips (when you hover over sidebar menu items for example).

public/css/sequenceserver.css
3lviend: Yeah, I'll update

3lviend: Updated.

yannickwurm:
This is great progress. 

There are also some (expected) formatting issues with e.g., heterogenous font sizes, spacing that doesn't match our previous spacing, and missing [+] things for showing/hiding the graphs

Screenshot 2024-08-05 at 10 14 53
Screenshot 2024-08-05 at 10 14 49

tadast:
Thanks! I'll have a proper look on Monday, `jest` tests are failing and will need to be fixed, are you familiar with it?

3lviend:
Hi @tadast! `jest` tests already fixed :raised_hands:

tadast:
After rebasing on master, there are some issues on the UI:


- [ ] option selection is orange, should be only orange on hover 
image

- [ ] JS error for the version checker when loading results page: 
image

- [ ] Sidebar overlays content when scrolling down (not when at the top of the page thought). Currently on master sidebar is not rendered on small screens at all. Perhaps we could just push it to be at the bottom instead, or make it slide out from the side when a sticky button is pressed, to not loose some features available only in the sidebar on small screens. Feel free to just hide it for now, perhaps we can handle it in a separate PR.

image

public/js/cloud_share_modal.js
```diff
@@ -117,25 +117,24 @@ export default class CloudShareModal extends React.Component {
 
     return(
       <form onSubmit={this.handleSubmit}>
-        <div className="form-group">
-          <label htmlFor="emailInput">Your Email Address</label>
+        <div className="px-6 mb-3">
+          <label htmlFor="emailInput" className="text-seqblue hover: text-orange cursor-pointer mb-0">Your Email Address</label>
```
tadast, line 121: I think you had a bad regexp/paste here and in many other places `hover: text-orange` -> `hover:text-orange`

public/js/databases_tree.js
```diff
@@ -76,22 +79,29 @@ export default class extends Databases {
...
                         {
                             this.renderDatabaseSearch(category)
                         }
-                        <button type='button' className={toggleClass} disabled={toggleDisabled} style={{ display: 'none' }}
-                            onClick={function () { this.handleToggle(toggleState, category); }.bind(this)}>
+                        <button
+                            type="button"
+                            className={toggleClass}
+                            disabled={toggleDisabled}
+                            style={{ display: 'none' }}
```
tadast, line 92: Do we still need to use inline styles here and in other places? Can we use tailwind classes instead?

views/layout.erb
```diff
@@ -10,82 +10,75 @@
   <meta name="description" content="Custom BLAST server provided by SequenceServer (https://sequenceserver.com)"/>
 
   <%= Rack::Csrf.metatag(env) %>
-  <!-- CSS -->
-  <% if settings.production? || settings.test? %>
-    <link rel="stylesheet" media="screen,print" type="text/css" href="css/sequenceserver.min.css?ver=<%= SequenceServer::VERSION %>"/>
-  <% else %>
-    <link rel="stylesheet" media="screen,print" type="text/css" href="css/sequenceserver.css"/>
-  <% end %>
+  <link rel="stylesheet" media="screen,print" type="text/css" href="css/app.min.css"/>
```
tadast, line 13: I'd say keep the `?ver=<%= SequenceServer::VERSION %>"` to bust browser cache between gem releases

views/layout.erb
```diff
@@ -10,82 +10,75 @@
   <meta name="description" content="Custom BLAST server provided by SequenceServer (https://sequenceserver.com)"/>
 
   <%= Rack::Csrf.metatag(env) %>
-  <!-- CSS -->
```
tadast, line 13: I wonder if we still need separate layouts for results and search. search_layout.erb was introduced when ONLY search page was converted to tailwind, perhaps they can be merged together again?

tadast (commented):
Very good progress, thanks!

- [ ] "👁️ Sequence" modal needs some padding and the close button position is broken

image

views/layout.erb
3lviend: Hi @tadast, Yes. We can merged them into one, but we have a different top nav style between search page and report page. do we want to keep both styles or choose one?

**search page**
image

**report page**
image

public/js/cloud_share_modal.js
3lviend: updated

public/js/databases_tree.js
3lviend: updated

views/layout.erb
3lviend: updated

views/layout.erb
3lviend: updated

3lviend:
Hi @tadast, yes currently the sidebar is hidden on mobile view

tadast:
Hey @3lviend thanks for progressing this and sorry for taking a while to respond.

I've rebased your commit against the latest changes and added a commit that uses the same layout for search and result pages. 

- [ ] This has highlighted an issue that UI still depends on styles defined in public/css/sequenceserver.css for many UI elements. That file also contains some unused CSS. Could you please go through react components and update them to use tailwind instead of relying on `sequenceserver.css`? If it is extremely laborious for some classes, move these styles over to public/css/app.css where they should be picked up and we can address them later.

- [ ] We still have bootstrap in package.json. Are we now able to remove it completely or do we still rely on it for anything?

- [ ] Could you please make all modals wider, maybe up to `max-w-screen-md`?

tadast:
Nice to see bootstrap gone! It's also great to see the new tooltips work consistently and have correct offset when scrolling, which was an issue on the main branch 🙌  thank you

 Some outstanding issues:

- [ ] Fix both rspec and jest tests please
- [ ] Share to cloud -> after sharing the modal UI breaks
image

- [ ] View sequence modal: the line numbers bar and the lines look like two separate rounded boxes. They should look like the same visual component
image

tadast:
- [ ] "Similar sequences" overlay the cloud share modal
image

- [ ] "View sequence" formatting still off
image

yannickwurm:
This is superb - huge thanks. 

Some requests: 

- [ ] when clicking submit, the "wheel" should spin. it is now static
Screenshot 2024-09-11 at 11 45 44

--- 

In the sidebar, the vertical spacing is somewhat inconsistent. Screenshot 2024-09-11 at 11 46 24

- [ ] Please add a small vertical space before (above) each bold section.
- [ ] Please ensure that vertical spacing between queries is similar to vertical spacing between Report types

---

 - [ ] When there are multiple hits, there is too much vertical spacing between hits
Screenshot 2024-09-11 at 11 50 31

---

  - [ ] Can we improve the horizontal alignment here? See the blue lines I drew. Ideally the left side of the alignment text is aligned to that blue line. And both the rightmost parts of the image text and of the alignment text are aligned to the blue line on the right
Screenshot 2024-09-11 at 11 55 56

---

 - [ ] here in a blasts result (but likely same for all) - when there are multiple HSPs a b, c... , have a bit more vertical space above the HSP title. 
Screenshot 2024-09-11 at 12 01 19

---


- [ ] ON that same one, it feels like the alignment text is LARGER than the title. This is resolved by changing font-size from .75 rem to .7 rem.  Screenshot 2024-09-11 at 12 04 31


---

  - [ ] When I click "Edit search", the query is in the right place, but the db selection has been lost

public/js/form.js
```diff
@@ -70,6 +74,18 @@ export class Form extends Component {
...
+
+                if (data['preSelectedDbs']) {
+                    const preselectData = data['preSelectedDbs'];
+                    $.each(preselectData, function(index, db) {
+                        const matchingCheckbox = checkboxDatabases.filter(`[value="${db.id}"]`);
+                        matchingCheckbox.prop('checked', true);
+                    });
+                }
+            }, 0);
+
```
tadast, line 101: Any reason we can't use React here? We've been trying to reduce dependency on jQuery in general as well as just using React for dom manipulation.

tailwind.config.js
```diff
@@ -11,6 +11,9 @@ module.exports = {
         seqblue: "#1B557A",
         seqorange: "#C74F13",
       },
+      fontSize: {
+        xs : ".7rem",
+      },
```
tadast: I don't think we should change global tailwind sizes. The reason is that if we use openly available components like Flowbite they will look different with tailwinds global config changed. 

Also, I believe @yannickwurm's main concern was inconsistent and too many different font sizes on the page, so just make sure we use font sizes consistently and if there's some non-tailwind CSS tweaking font sizes via global CSS rules, address that instead

public/js/hsp.js
```diff
@@ -35,12 +35,12 @@ export default class HSP extends React.Component {
...
     return (
       <div
-        className="hsp pt-px pr-0 pl-5 pb-0.5"
+        className="hsp pt-px pb-5"
         id={this.domID()}
         ref={this.hspRef}
         data-parent-hit={this.hitDOM_ID()}
       >
-        <p className="m-0 p-0 rounded-none border-0 bg-inherit whitespace-pre-wrap break-keep text-sm text-gray-700 font-semibold tracking-tighter">
+        <p className="m-0 p-0 rounded-none border-0 bg-inherit whitespace-pre-wrap break-keep text-sm text-neutral-500 font-semibold tracking-[0.7px]">
```
tadast, line 43: `tracking-[0.7px]` - same here, I'd rather stick to standard tailwind sizings even though it looks different from the previous version (current main branch). As long as the font sizes are consistent on the same page I'd prefer to stick to Tailwind conventions

yannickwurm:
- [ ] vertical spacing is still too much here
 - [ ] alignment doesn't seem to be wrapping in the right places?  (I see this in multiple browsers)
 
Screenshot 2024-09-20 at 14 38 42

Screenshot 2024-09-20 at 14 41 48

yannickwurm:
I'm happy with this - thank you

PR #786, opened 2024-08-27
Update to use d3 v7

joko3ono:
update d3 to use the latest version 7 to solve
d3-color vulnerable to ReDoS - https://github.com/advisories/GHSA-36jr-mh4h-2g58

After the update, CircosJS is no longer functional. I attempted to resolve this by updating the library, but it made the code difficult to maintain.

There are more updates to this branch, the updates are :
1. visualization helpers still throw an error
2. update histogram to work with latest d3


tadast (commented):
@joko3ono nice, thanks. 
Am I right in thinking that circosjs.js has some backported features/patches that are custom to sequenceserver? If that's the case, would you please mind add some comments at the top of that file explaining what the patches do and why we need them, so that next time someone looks at it they know the full context and the reasons why that code exists?

Otherwise looks good 🙌

tadast:
image

🕺 🕺 🕺

Issue #485, opened 2020-08-25
Improve FASTA Download


enuggetry:
Currently, FASTA download can take a long time to process before downloading takes place.  This is understandable given the use of blastdbcmd to extract the FASTA data.  However, during the processing, the app is hung while waiting for the processing.
I'd like to see it display a progress indicator while it is waiting, as it can take several minutes to process.  Or, it can somehow do this processing in the background.  (not a high priority.  Just a suggestion)

yeban:
I agree the UI shouldn't hang while waiting for the download and should probably show a progress indicator as the download can indeed take a while. I would also like any error during download to be displayed in the standard error modal instead of navigating to a new page.

I think this is a kind of enhancement that can be included in a point release after 2.0 stable.

lukaszsobala:
Hello,

I don't know how exactly sequenceserver uses blast commands to download the full FASTA files, but this seems to be possible to improve. From what I understand, right now it generates a single command line containing all the sequence IDs it wants, and it sometimes fails if the command is too long (correct me if I am wrong). I cannot find out how many it is in the current version (sequenceserver 3.1.2), I tried up to 3500 hits.

What do you think about changing this behaviour in favour of using `blastdbcmd` (which supposedly can be used with multiple databases) and `-entry_batch` with a list of identifiers to download? Right now (at least for me) it is an extra step to download all the results if there are too many to download from the sequenceserver web gui.

Or one could think of running `blastdbcmd` multiple times in parallel (for example, sequentially akin to GNU `parallel` using as many threads as the user defined) as a fallback if the number of sequences is too large for a single download, then combining the output.

If this is impossible to introduce, one thing that could be improved is doing something like `sort|uniq` on the FASTA sequence identifiers. This way, in case of many query sequences, only the non-redundant hits would be downloaded. The changed order of hits would not matter when downloading all of them in such a case, because sequenceserver returns a single multi-FASTA file anyway.

yannickwurm:
Hey @lukaszsobala 
We actually should have  resolved this last week: https://github.com/wurmlab/sequenceserver/pull/754
This should be in 3.1.1 and 3.1.2 and I think addresses your issue?

Are you able to test with the newer version?

Cheers
Yannick

lukaszsobala:
@yannickwurm 

[...]

PR #790, opened 2024-09-12
Redudancy removal #485

joko3ono:
I do test using sending 2 identiq query, it will give me redundant results, but when I do download it giving me this error message on fasta file

```
# ERROR: incorrect number of sequences found.
# You requested 14 sequence(s) with the following
# identifiers:
#   SI2.2.0_06267, ref|XM_039451179.1|, ref|XM_039451177.1|, ref|XM_039451188.1|, ref|XM_039451192.1|, ref|XM_011159896.3|, ref|XM_039451197.1|, SI2.2.0_06267, ref|XM_039451179.1|, ref|XM_039451177.1|, ref|XM_039451188.1|, ref|XM_039451192.1|, ref|XM_011159896.3|, ref|XM_039451197.1|
# from the following databases:
#   Hereis assembly one 2022-04 , sequenceserver-18 hits
# but we found 7 sequence(s).
# 
# This is likley due to a problem with how databases are formatted.
# Please share this text with the person managing this website.
# 
# If you are the admin and are confident that your databases are
# correctly formatted, you have likely encountered a weird bug.
# In this case, please raise an issue at:
# https://github.com/wurmlab/sequenceserver/issues
# 
# If any sequences were retrieved, you can find them below
# (but some may be incorrect, so be careful!)
```

### DIFF
--- a/lib/sequenceserver/blast/job.rb
+++ b/lib/sequenceserver/blast/job.rb
@@ -7,6 +7,7 @@ module SequenceServer
   module BLAST
     # Extends SequenceServer::Job to describe a BLAST job.
     class Job < Job
+      # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       def initialize(params)
         if params.key?(:xml)
           super do
@@ -22,7 +23,7 @@ module SequenceServer
           validate params
           super do
             @method = params[:method]
-            @query = params[:sequence]
+            @query = distinct_sequences(params[:sequence])
             @qfile     = store('query.fa', params[:sequence])
             @databases = Database[params[:databases]]
             @advanced  = params[:advanced].to_s.strip
@@ -35,6 +36,7 @@ module SequenceServer
           end
         end
       end
+      # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
 
       # :nodoc:
       # Attributes used by us - should be considered private.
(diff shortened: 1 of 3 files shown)


tadast:
So this passes all sequence ids to blast and only when it's done, looks for redundant sequences. @yannickwurm am I right in thinking that we could do this check earlier and eliminate redundant sequence **IDs** before sending them to blast?

If that's the case, it would be easiest to do it as early as possible - in the frontend where we're compiling sequence IDs to post to the backend.

@joko3ono it would also be nice to get some rspec tests for this change.

yannickwurm:
Yes, Tadas as you suggest it would be better to Remove redundancy beforehand. 
Once we’ve done that, we should not remove it done second time

joko3ono:
@tadast logic changed just before blast, if there is duplicate query, we will only take the first one and ignore the last one

tadast:
Ok this is interesting, you're actually looking at the query sequence **values**, not **identifiers**, and removing duplicates based on value uniqueness. 

Originally I suggested making the list of requested sequence IDs unique before the user downloads them [here](https://github.com/wurmlab/sequenceserver/blob/57c4fc22e1e5d7d85c81ba727133ed1c97da27b1/lib/sequenceserver/routes.rb#L159) or even [here](https://github.com/wurmlab/sequenceserver/blob/57c4fc22e1e5d7d85c81ba727133ed1c97da27b1/public/js/download_fasta.js#L8) to address the immediate issue, but this takes it a step further and eliminates duplicates before running BLAST.

I feel like the user could still make an error and have something like:
```
> Query_1
ACGT
> Query_1
AGGT
```

i.e. the sequences are unique, but IDs are mistakenly duplicated.

Similarly, the solution currently proposed ignores duplicate sequences even if their IDs are unique, only considering the first one.

I'd imagine there are performance implications for large queries too when comparing sequences in ruby.

@yannickwurm @djleite what do you think about this?

yannickwurm:
We should indeed deduplicate based on sequence identifiers. (those we have in the blast report). 

We should do this prior to any retrieval.

tadast:
@joko3ono sounds like https://github.com/wurmlab/sequenceserver/blob/57c4fc22e1e5d7d85c81ba727133ed1c97da27b1/lib/sequenceserver/routes.rb#L159 is the right place then. You can discard this PR or replace the commits, whichever is easier.

joko3ono:
@tadast I will close this PR
use this PR instead #803

Issue #416, opened 2019-10-08
download taxdb and import xml options


yeban:
In bin/sequenceserver —download-taxdb and import xml option can be triggered before calling SequenceServer.init

Basically, going through sequenceserver’s startup routine (where it checks for blast binaries and databases) is unnecessary for these two options

yeban:
It's not important to track it here. There are many examples of 'refactoring required' in the codebase that are more important than this one. We can consider them together in due time.

PR #797, opened 2024-09-26
Fix: bad unit labels

joko3ono:
Fix unit labels from

`1k bp` or `12M aa`

to 

`1 kbp` or `12 Maa`

### DIFF
--- a/public/js/visualisation_helpers.js
+++ b/public/js/visualisation_helpers.js
@@ -43,7 +43,7 @@ export function tick_formatter(scale, seq_type) {
 
     return function (d) {
         return `${prefix(d)}${suffixes[seq_type]}`
-            .replace(/([0-9])([a-zA-Z])/g, '$1 $2');
+            .replace(/([a-zA-Z]+)/, ' $1')
     };
 }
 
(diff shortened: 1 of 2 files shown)


public/js/visualisation_helpers.js
```diff
@@ -42,7 +42,8 @@ export function tick_formatter(scale, seq_type) {
     var suffixes = {amino_acid: 'aa', nucleic_acid: 'bp'};
 
     return function (d) {
-        return `${prefix(d)} ${suffixes[seq_type]}`;
+        return `${prefix(d)}${suffixes[seq_type]}`
+            .replace(/([0-9])([a-zA-Z])/g, '$1 $2');
```
tadast, line 46: Minor, feel free to ignore: I think this would be simpler and achieve the same result?
```suggestion
            .replace(/([a-zA-Z]+)/, ' $1')
```

tadast (approved):

djleite:
@tadast @joko3ono can this be merged to main? and we close [#416](https://github.com/wurmlab/sequenceserver-cloud/issues/416)

public/js/visualisation_helpers.js
joko3ono: @tadast yes you are right, code updated and replace code also added to length distribution

PR #803, opened 2024-10-02
deduplicate identifier on download all fasta

### LINKED ISSUE #485: Improve FASTA Download

joko3ono:
resolve #485


tadast (commented):
LGTM, please add an RSpec test :)

Issue #802, opened 2024-10-01
fix sidebar and FASTA ID wrapping


djleite:
### Original
![image](https://github.com/user-attachments/assets/af0c6bcf-d8a7-4b08-8290-f6306946446a)
### Update
![image](https://github.com/user-attachments/assets/d66653f7-c161-4123-9390-d31f59b92bc3)
 
- [ ] The siderbar options for selected hits should be
      grey: when no sequences are selected
      blue: when sequences are selected
- [ ] The "Share results" section has too much white spacing around the options

### Original
![image](https://github.com/user-attachments/assets/ba66388e-7203-4b9f-8f50-2ea22ef91a12)
### Update
![image](https://github.com/user-attachments/assets/cdc230b8-846d-4649-b05b-e9bfa393a66c)

- [ ] The FASTA IDs need to wrap so they do not extend onto new lines

djleite:
@3lviend please can you take a look at these?

PR #804, opened 2024-10-03
Fix sidebar and FASTA ID wrapping

3lviend:
issue [#802](https://github.com/wurmlab/sequenceserver/issues/802)

- Sidebar option selected 
image

- Share results
image


public/js/hits.js
```diff
@@ -210,27 +210,27 @@ class Hits extends Component {
         // Highlight selected hit and enable 'Download FASTA/Alignment of
         // selected' links.
         if (checkbox.is(':checked')) {
-            $hit.addClass('glow');
-            $hit.next('.hsp').addClass('glow');
+            $hit.addClass('border-l-2 border-seqorange pl-1');
+            $hit.next('.hsp').addClass('border-l-2 border-seqorange pl-1');
```
tadast, line 214: I like the highlight, but it also shifts both blocks to the right. I think it would be nice to offset the added border by a negative left margin or something so that checking the box does not move elements visually

tadast (commented):
Code looks good to me, just one change request

public/js/hits.js
3lviend: updated.

tadast (approved):
Code LGTM, thanks Elvin. 

@djleite would you like to test to confirm this fixes the issue you've raised?

djleite:
@3lviend - thank you.

I see two issues still

1. it seems that when I blast nucleotide to protein I don't get any option to download FASTA files.

image

2. there is no grey boarder appearing when moused over.

https://www.loom.com/share/49ebf76b3ffe44899272c6fdb653f03b?sid=a2738378-f7b2-4e0e-84e7-b14171e29a43

tadast:
> it seems that when I blast nucleotide to protein I don't get any option to download FASTA files.

Download links are disabled when you query against database that was formatted without parse_seqids, I did a test with a correctly formatted DB and the links show up. Shout if that's not right @djleite

PR #806, opened 2024-10-05
Fix: reformat long seqs

joko3ono:
resolve wurmlab/sequenceserver-cloud#415


tadast (commented):
@joko3ono thanks for looking into this. Could you please add an explanation of what was the issue and how your change fixes the issue to the commit message, and also add some jest unit tests to demonstrate how it works + make sure we don't break it in future?

It's a bit hard to review the changes without context and learning what the bug is from scratch would be very time consuming as I'd have to come up with multiple test scenarios and execute them 😬

djleite:
things look almost ok 

however, there needs to be correct rounding up of the final sequence length. See below that:

- 3,999,920 bp should be 4 Mbp
- 799,920 bp should be 800 Kbp

image
image

### However...
If we want to show a greater level of precision then we can show to at least 1 decimal place for any value that has units kbp, Mbp and Gbp.

ie
divide these numbers by 1000 and round to the nearest decimal place
- 19,299 bp = 19.3 kbp
- 19,990 bp = 20.0 kbp

divide these numbers by 1000000 and round to the nearest decimal place
- 199,929,999 bp = 199.9 Mbp
- 199,990,000 bp = 200.0 Mbp

tadast:
@joko3ono also please make sure you add tests for the rounding so that it's not broken in future

public/js/visualisation_helpers.js
```diff
@@ -31,22 +31,45 @@ export function getPrefix(str) {
...
+    const [numericPart, suffix] = extractSuffix(number);
+    const integerPart = Math.floor(numericPart);
+    const fractionalPart = parseFloat((numericPart - integerPart).toFixed(2));
+
+    // when fractionalPart more than the threshold round up
+    // example:
+    //  threshold 0.95
+    //  2.95 -> 3
+    //  2.94 -> 2.9
+    const formatted = fractionalPart >= threshold ? integerPart + 1 : Math.floor(numericPart * 10) / 10;
```
tadast, line 68: Do we need this custom rounding logic threshold? Am I right in thinking that we could instead round up to 1 decimal place, then remove ".0" at the end if it's a whole number? I think it would achieve the same result with a much easier to understand and simpler code, or am I missing something?

public/js/tests/visualisation_helpers.spec.js
```diff
@@ -0,0 +1,28 @@
+// Import the necessary functions
+import { tick_formatter, formatNumberUnits } from '../visualisation_helpers';
```
tadast, line 2: Thanks for adding the tests!

public/js/visualisation_helpers.js
joko3ono: @tadast the threshold based on suggested logic from Daniel

> divide these numbers by 1000 and round to the nearest decimal place
> 
> 19,299 bp = 19.3 kbp
> 19,990 bp = 20.0 kbp
> divide these numbers by 1000000 and round to the nearest decimal place
> 
> 199,929,999 bp = 199.9 Mbp
> 199,990,000 bp = 200.0 Mbp

joko3ono: @tadast I updated the code, it's more simple now

public/js/tests/visualisation_helpers.spec.js
joko3ono: test also updated

tadast (approved):
Thanks @joko3ono, the code looks good now. 

@yannickwurm do you want to test this or happy to merge?

Issue #738, opened 2024-04-16
Advanced params help modal should be bigger - take up more of the page so we don't need to scroll as much


yannickwurm:

yannickwurm:
![Screenshot 2024-07-26 at 15 09 20](https://github.com/user-attachments/assets/8a327455-2826-4e35-b2d5-fcd764df1c23)

tadast:
Let's fix it, but do it in a proper way - at the moment the modals are built like it's 2010. Browsers [have a native dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#open) nowadays that is much less HTML and much less javascript when you use it. Let's update all modals to use that - either directly or via an off-the-shelf react component that uses `<dialog>` as well as review the sizes of all modals and make them bigger where they have a lot of content to display.

PR #808, opened 2024-10-08
Convert modal

3lviend:
Issue [#738](https://github.com/wurmlab/sequenceserver/issues/738)

Convert modal using dialog:
[dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog)

modal update:
image
image
image
image


tadast:
image

Looks like the tooltips for sequence viewer modal are no longer working. Above is an example from main branch, on this branch the position tooltip is not shown and I can see the DOM has `undefined` strings as IDs.

Other than that looks good to me, the modals are now better sized and I can close them by hitting `esc` 🙌

PR #810, opened 2024-10-09
Add CSRF_PROTECTED env param

wangqian0306:
I've never written Ruby, so this might not be the best solution. I tried to introduce this variable using a configuration file but had some scope issues, so I solved this problem by using environment variables in other parts of this file.

### DIFF
--- a/lib/sequenceserver/routes.rb
+++ b/lib/sequenceserver/routes.rb
@@ -65,8 +65,8 @@ module SequenceServer
         key: 'rack.session.sequenceserver',
         secret: ENV.fetch('SESSION_SECRET') { SecureRandom.alphanumeric(64) }
       )
-      csrf_protected = ENV.fetch('CSRF_PROTECTED', 'true')
-      if csrf_protected == 'true'
+      
+      unless ENV['SKIP_CSRF_PROTECTION'] == 'true'
         use Rack::Csrf, raise: true, skip: ['POST:/cloud_share']
       end
     end
--- a/lib/sequenceserver/routes.rb
+++ b/lib/sequenceserver/routes.rb
@@ -65,10 +65,8 @@ module SequenceServer
         key: 'rack.session.sequenceserver',
         secret: ENV.fetch('SESSION_SECRET') { SecureRandom.alphanumeric(64) }
       )
-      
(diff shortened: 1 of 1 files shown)


tadast (changes requested):
Thanks for this!

However, I'd like a couple of changes before accepting this please:

1) By default CSRF protection should be true, so users who do not change any configuration should have it enabled
2) We fetch config from a configuration file, I'd like to keep that consistent.

So if you add `skip_csrf_protection` with a default of `false` in [default config](https://github.com/wurmlab/sequenceserver/blob/master/lib/sequenceserver/config.rb#L95) and then wrap the  `use Rack::Csrf` inside `unless SequenceServer.config[:skip_csrf_protection]` it should satisfy both of my requirements.

wangqian0306:
@tadast Thanks for your guidance. My English is too poor to be clear in the description. I did the same thing at first, like `SequenceServer.config[:frame_options]`. But maybe due to the initialization order or the scope, The program can't read the `SequenceServer.config` method properly here, it can't start. The error message like this:

image

Then I saw in the code here that the program also used environment variables, so I adopted this approach.

 Hope you can provide further explanation or guidance.

lib/sequenceserver/routes.rb
```diff
@@ -65,8 +65,10 @@ class Routes < Sinatra::Base
         key: 'rack.session.sequenceserver',
         secret: ENV.fetch('SESSION_SECRET') { SecureRandom.alphanumeric(64) }
       )
-
-      use Rack::Csrf, raise: true, skip: ['POST:/cloud_share']
+      csrf_protected = ENV.fetch('CSRF_PROTECTED', 'true')
+      if csrf_protected == 'true'
```
tadast, line 69: Ok if config is not loaded at this point let's just invert the logic in your originally proposed change so that CSRF protection is on by default:

```suggestion
      unless ENV['SKIP_CSRF_PROTECtION'] == 'true'
```

wangqian0306:
@tadast Thanks for your help, this looks much better.

tadast (approved):
Thanks @wangqian0306, happy to accept this change now. Congrats on your first ruby open source contribution 🙌 

I'll merge this and look in to updating API docs with this temporary workaround as well as releasing a new version of the gem.

Issue #815, opened 2024-10-25
Pairwise alignments are misformatted in PDF output


yannickwurm:
When you print to PDF in Safari, it generally looks great. But some of the formatting gets broken:

<img width="415" alt="Screenshot 2024-10-25 at 17 01 44" src="https://github.com/user-attachments/assets/ada42a92-7636-434b-a21f-66e2c9358c44">

(WE need to downsize the font or wrap to next line earlier)

PR #816, opened 2024-11-12
Add print layout with ability to hide

joko3ono:
This PR aims to resolve issue #815.

I've added a print layout to make the search results more readable when printed. Here is the result for printing all pages.

[print-all.pdf](https://github.com/user-attachments/files/17711111/print-all.pdf)
Screenshot 2024-11-12 at 11 27 15

With the ability to fold certain content, I used this feature to completely hide folded sections in print. Here’s an example where the "Graphical Overview" section is folded.

[exclude-circos.pdf](https://github.com/user-attachments/files/17711107/exclude-circos.pdf)
Screenshot 2024-11-12 at 11 27 39


yannickwurm:
Thank you Joko - what would happen if an American tries printing? (they are typically US Letter paper size users)Screenshot 2024-11-12 at 14 08 03

public/js/collapse_preferences.js
```diff
@@ -1,13 +1,20 @@
 export default class CollapsePreferences {
-    constructor(component) {
+    constructor(component, reactFunction = false) {
```
tadast, line 2: This change feels wrong, you are essentially changing the function interface by passing in `reactFunction = true`. We don't use `CollapsePreferences` in many places, would it be fairly easy to update the other components to `useState` and `useEffect` so that `CollapsePreferences` interface remains the same?

public/js/grapher.js
```diff
@@ -1,131 +1,129 @@
...
+        const name = Graph.name(props);
+        const [width, setWidth] = useState(window.innerWidth);
+        const [collapsed, setCollapsed] = useState(false);
+        const svgContainerRef = createRef();
+        let graph = null;
 
-        graphId() {
-            return Graph.graphId(this.props);
-        }
+        const graphId = () => "testing"
```
tadast, line 22: I guess the PR is still WIP?

public/js/collapse_preferences.js
joko3ono: I did use the `collapsePreferences` function in a React functional component before (in the `ReportHeader`). Now, the collapse preferences have reverted to their initial state.

tadast (approved):
Looks good, thanks

Issue #784, opened 2024-08-15
BLAST button not working with sequenceserver 3.1.3 and tree widget enabled


lukaszsobala:
Hello,

The issue is explained in the description. When the tree widget is disabled, the BLAST button does work. I have tested this on 2 separate devices. Also, I have tested in both Firefox and Chromium.

The bug seems to be related to the predefined command line options, which don't seem to be loading at all. They are quite slow to load in 3.1.2 already...

lukaszsobala:
Hi Yannick,

I am sorry but I have no clue about how to debug javascript!

When you paste a sequence here, for example: https:///lukprot.hirszfeld.pl/ the options I predefined take 5-10 seconds to load up on 3.1.2 when you select the databases. This might be because there are so many databases in the tree. On 3.1.3 this does not seem to work at all (even with 1 database selected when the tree widget is active).

Cheers,
Łukasz

kuntzagk:
I have run into the same problem here. When examining the delivered javascript it looks like the "query_stats.js" delivered is much smaller then in 3.1.2. 
It is identical to the code checked into GH while on 3.1.2 it contains other contain (when loaded in browser, code on disk is identical)
So I guess some kind of magic code replacement is not happening. 

hth

jasongallant:
+1  I had a user report in on this issue from https://www.efishgenomics.com/blast_search/ with the same issue.  I disabled tree for the time being, but it sure is much messier looking.  Any ideas what might be behind this?

Thanks in advance

yannickwurm:
Hi folks - there have been several changes in the underlying code and process. 

I'm unsure as to why this broke. A likely contributor is our migrating from using a long deprecated version of bootstrap to using tailwind, and updating a bunch of deprecated javascript libraries. Another possible contributor are the increased cross-sight forgery protections. 

We'd be most eager for contributions that look into this. The likely best place to begin any fixes is on what's currently in master. Though a first step is likely here https://github.com/wurmlab/sequenceserver/pull/799

PR #799, opened 2024-09-30
[WIP/Discussion] Trigger a "change" event when tree.js DB is selected

tadast:

PR #821, opened 2024-11-21
Fix: Button remains disabled despite selecting a database in the tree widget

joko3ono:
Visible Issue:  
- Nodes should not be unchecked when opening the tree widget.


tadast (approved):
Gave this a little test and all looks good.

~@joko3ono did you consider the latest tree.js react version? Not saying you should've used it, just curious if you evaluated it and what effort would it take to use it.~ NVM, saw a comment on https://github.com/wurmlab/sequenceserver/pull/799

tadast:
fixes #784

PR #828, opened 2025-05-21
Fix #826: top-align identifier

### LINKED ISSUE #826: Identifier is bottom-aligned while it should be topaligned
Image

Good-soup1:
I have emulated on different devices through the dev tools, also tested on microsoft edge as well as chrome. Also tested on an extra monitor which has different dimensions and it seems to work consistently.   
826_top-align


yannickwurm:
Thanks!
Please include screenshots - did you check it works across devices, and screen sizes and browsers?

tadast:
Thanks @Good-soup1 🙌

